PR #2, opened 2024-05-29
Support serving static files

zacharygolba:
Adds support for serving static files with the `via-serve-static` crate by making the following changes:

- Adds a `Body` enum to support streaming responses.
- Adds a `param` method to `Location` to enable middleware authors to use dynamic path parameters without having use macros.
- Implements `Respond` for `tokio::fs::File`. A similar approach can be used  in the future if we need to implement respond for other types that implement `AsyncRead`.
- Creates a `via-serve-static` crate with the actual implementation of the static file server.

### DIFF
--- a/crates/via-serve-static/src/lib.rs
+++ b/crates/via-serve-static/src/lib.rs
@@ -1,45 +1,67 @@
 use std::path::{Component, Path, PathBuf};
 use tokio::fs::{self, File};
-use via::prelude::*;
+use via::{prelude::*, routing::Location};
 
-pub struct ServeStatic {
-    root: PathBuf,
+pub struct ServeStatic<'a> {
+    location: Location<'a>,
 }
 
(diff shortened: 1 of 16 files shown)


crates/via-serve-static/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+                .first_or_octet_stream()
+                .to_string()
+                .parse()?,
+        );
+
+        Ok(response)
+    }
+}
+
+fn normalize_path(path: &Path) -> PathBuf {
```
zacharygolba, line 46: Copied from: https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61

zacharygolba:
Closing this PR in favor of #3.

PR #3, opened 2024-06-07
Re-write and re-imaging of via

zacharygolba:
# Background

I started this project about four years ago as an experiment in exploring different code reuse patterns and separating concerns in Rust. At the time, async Rust was still in its infancy, and this project helped me understand the performance characteristics and constraints associated with concurrent programming in Rust. After putting together a few examples inspired by other web application frameworks with a strong OO influence, I eventually shelved the project for a couple of years.

Fast forward to June of this year. I found myself with some spare time and decided to dive back into Rust. What better way to reacquaint myself with the language than by revisiting an old project, right? Returning to the project with fresh eyes, I realized my original approach to thinking about Rust code was not very... rusty. This led me back to the drawing board to reassess my goals. With a background in the agency world and experience in both backend and frontend projects, I wanted to create something I would have appreciated in past projects. For better or worse, this brought me back to thinking about JavaScript.

JavaScript has been a popular choice for many high-profile clients at the agency I previously worked for. Being a multi-paradigm language, JavaScript allows programmers to express themselves in many ways. Combining its good parts with a relatively quick learning curve, it becomes a reliable choice for companies aiming for productivity. This got me thinking: since Rust is also a multi-paradigm language with an expressive type system and powerful features, is there a way to combine simplicity, power, and expressiveness to make Rust the best choice for backend server development?

Async Rust has evolved significantly over the past four years. Like JavaScript, it has both its strengths and weaknesses. Fortunately, many great frameworks now make it enjoyable to be a Rust programmer. If your team knows what they're doing, it's hard to go wrong with the available frameworks today. With that said, I envisioned a framework that caters to both Rust experts and beginners. Here is what I came up with.

# Principles

At the start of any project, it's essential to define the problem you're solving and decide on the trade-offs you're willing to make. We've seen time and again that file systems, databases, and external services often become the bottlenecks in a web application backend. However, we're writing Rust for a reason, so let's not be wasteful with CPU cycles or memory where it matters. But let's also be practical and think about the people writing the code rather than just the computer running it.

Hyper is a state-of-the-art HTTP library that offers excellent performance out of the box. Like many other frameworks, starting with a solid foundation, such as a fast and correct HTTP implementation, is crucial. However, what happens next can make or break the rest of the stack.

Consider the following principles like layers of a cake, each representing a different level of abstraction. Starting from the bottom up. Except for the section on Security—it's mentioned last but it's not the top layer. It's more like the plate that the cake is served on.

## Fast and Flexible Routing

One hallmark of a great web application backend is the separation of concerns. When onboarding to a new project, a great way to start is by examining all the routes available in an application. URL paths and routes often form a tree-like structure that indicates the separation of concerns for a particular web server. For example, routes prefixed with `/api` will likely return a JSON response. With that in mind, let's use the path as a starting point for code reuse. Since the router I started working on four years ago could be considered a prefix-tree, we were already off to a good start.

### Philosophy

An engineer should be able to open a file and see every possible route that can serve a request. Co-locating associated middleware for a route and its descendants alongside the HTTP method a route accepts makes it easy to see what happens when someone visits the corresponding URL. Let's not introduce any ambiguity here. In my opinion, it's more important to see the HTTP method at the route definition rather than the function. Additionally, it's helpful to know that the function signature for any given middleware or responder is `async fn _(request: Request, next: Next) -> Result<Response>`. Therefore, I decided to remove the codegen crate to rule out the use of proc-macros to define routes.

Many frameworks, such as Tide, follow this philosophy, and you can see Tide's influence on the public API of the router. However, there's one crucial difference in how routing works in via compared to Tide. Via supports the concept of partial route matches by leveraging the tree-like data structure of our routes. The benefit of this feature is that we can use the URL path as a 1-to-1 mapping of the separation of concerns across routes and resources in a via application. For example, suppose a new JIRA ticket is automatically created from webhooks to an error reporting service with the title `JSON.parse: unexpected character at line 1 column 1 of the JSON data`. This is likely caused by an error response returned from an endpoint that should respond with JSON, but in case of an error, a plaintext response is sent to the client. With via, you can include middleware that runs unconditionally anytime the URL path partially matches a route. This means you can apply error-handling logic, compression, authentication for any middleware or responder downstream from `/api` without having to repeat yourself or introduce some higher-order function that eventually becomes the source of every heisenbug in your app. The benefit of defining routes this way can be seen in the [blog-api](https://github.com/zacharygolba/via/blob/feat-multi-route-match-with-slab/docs/examples/blog-api/src/main.rs) example.

```rust
let mut api = app.at("/api");

// Apply specific error handling logic to the /api namespace.
api.include(ErrorBoundary::map(|mut error| {
    use diesel::result::Error as DieselError;

    if let Some(DieselError::NotFound) = error.source().downcast_ref() {
        // The error occurred because a record was not found in the
        // database, set the status to 404 Not Found.
        *error.status_mut() = StatusCode::NOT_FOUND;
    }

    // Return the error with the response format of JSON.
    error.json()
}));

api.at("/posts").scope(|posts| {
    use api::posts;

    posts.include(posts::authenticate);

    posts.respond(via::get(posts::index));
    posts.respond(via::post(posts::create));

    posts.at("/:id").scope(|post| {
        post.respond(via::get(posts::show));
        post.respond(via::patch(posts::update));
        post.respond(via::delete(posts::destroy));
    });
});

api.at("/users").scope(|users| {
    use api::users;

    users.respond(via::get(users::index));
    users.respond(via::post(users::create));

    users.at("/:id").scope(|user| {
        user.respond(via::get(users::show));
        user.respond(via::patch(users::update));
        user.respond(via::delete(users::destroy));
    });
});
```

### Performance

Now that we've defined how we want the router to work, the next step was to build it and make it fast. The result of the time spent optimizing the router when I started this rewrite a month ago can be seen by running [the benchmark in via-router](https://github.com/zacharygolba/via/blob/feat-multi-route-match-with-slab/crates/via-router/benches/bench.rs) on this branch. On my machine, it takes roughly `135ns` per iteration to find all matching routes for a path with nested resources in a router with 100 routes. This is better than I anticipated and fast enough to allow for some compromises higher up in the stack.

## Pragmatic Trade-Offs

We all have fought with the borrow checker and know how it feels after a long day of helpful yet despairing compiler errors. Often, the solution to these errors involves re-imagining and re-ordering the sequence of events in a function in a way that makes sense to a computer. This is an important part of being a Rust programmer. Nobody denies that. However, it's a hard conversation to have with your manager when you spend all day fighting the borrow checker, only to find

 out that the implementation using borrowed data is slower than the version you initially wrote using `Arc`. It's also no secret that using borrows unnecessarily only makes it more difficult to onboard a Rust beginner. Still, this is a complex subject in the language, and we shouldn't hide the parts of Rust that one will inevitably have to use.

### The Middle Ground

With via, I think it's important to find the intersection of performance and ease of use. Therefore, when it comes to data in a request, we treat performance requirements as a top priority. This means that when you work with path and query parameters from a `via::Request`, we return references (with the underlying type `&str`) that truly reference the data in the URL path. When you read the request body in a particular responder, the data is moved out of the request rather than copied to avoid wasting resources downstream. Above all, query parameters and other types of data related to a request (cookies, etc.) are always [parsed and stored lazily](https://github.com/zacharygolba/via/blob/feat-multi-route-match-with-slab/src/request/request.rs#L61-L64) so the query parser code does not run unless you actually use query parameters.

However, we don't want you to stay late at work unless you really, really want to. Therefore, we do some things that might be considered controversial by some in the Rust community (I'm half-joking here). We wrap middleware and responders added to a route in an `Arc` and use an owned reference to call the middleware when we unwind the stack because I like using the termination operator on line 1 of the function I'm writing. In all seriousness, I think it's important for productivity and onboarding to allow for `request` and `next` to move into the future returned from middleware. I believe the performance implications of doing so are acceptable for anyone comfortable using a framework rather than writing a hyper service from scratch for their next project. We also return a reference to an `Arc` of the global state returned from [`Request::state`](https://github.com/zacharygolba/via/blob/feat-multi-route-match-with-slab/src/request/request.rs#L77) so you can clone it as much as you want and tell your boss that we told you to do it!

## Unix-Like

When I initially started this four years ago, I wanted to provide a batteries-included option (similar to Django and Rails) but for Rust. However, I think it's important to support an ecosystem of libraries as unique as the community for which this framework was built. It's fun to write your own CORS middleware or static server, and we don't want to take that away from you. Via is meant to be a thin layer on top of hyper, giving you control over the important decisions. We'll provide a few "official" crates, such as `via-serve-static` introduced in this pull request, but we won't require you to use it or make it a part of our core API. Write your own if that makes you happy. That's what I told myself when I was looking at Rust web frameworks, and had I not done that, I would have been so bored last month. Competition is good for everyone, and I want to see if someone else will write an auth crate so I can deny culpability if need be. We'll also feature flag the stuff you expect, such as JSON support via `serde`, and never include default features.

** *via-serve-static is very, very fast and secure, and is a great choice for serving static files, but you can still write your own if you want to.*

## Security

Rust allows us to write code that is verified to be correct at compile time. I believe it is our responsibility to uphold the guarantees advertised by the language and not introduce unsafe blocks for a slight or even substantial performance improvement. I'm okay with being second best (or even in the top 5) if that means it's easier for me to sleep at night. I know what it's like to be a victim, and I love that Rust prioritizes safety and security.

The only place you should see `unsafe` in the source code is for Pin projections, as I believe it is important for us to write our own pin projections to verify their correctness as a community.

# Epilogue

Thank you for taking the time to read this description. I know test coverage can be improved, and that's a priority of mine, but I appreciate your time and effort in reviewing this substantial PR. If anyone can confirm the existence of and/or pinpoint the source of a small but potential memory leak, I promise you will be the first to receive a via t-shirt if this framework becomes popular. I welcome your constructive feedback and insights. Let's work together to make this project reach the potential that I believe it has!

You can find more traditional examples in this work-in-progress branch: https://github.com/zacharygolba/via/tree/refactor-simplified-examples/docs/examples.

Stay tuned as I plan on adding some issues for features that didn't make it into this pull request over the next week.

---

**Edit:** From what I can tell, the memory leak in via-router has been fixed. There may be improvements to be made with regards to memory management but memory usage should be stable at this point.

---

## Short-Term and Medium-Term Roadmap

**Blocking the first release:**

- [X] Refactor the `Request` type and `Error` struct to prevent the need to pass around a channel or `EventListener`. ([d714bc7](https://github.com/zacharygolba/via/pull/11/commits/d714bc7a5be2b2a16415da3f2fcba737e846d9b0))
- [X] Use a semaphore or some other mechanism to prevent the server from crashing if there are too many open file descriptors ([2a6be5a](https://github.com/zacharygolba/via/pull/11/commits/2a6be5aebe519d58c28f5f0677ad06e155b92fa4))
- [X] Support graceful shutdown. ([a8be8de](https://github.com/zacharygolba/via/pull/11/commits/a8be8deed3b218e11d3300e7ff4789448200df6b))
- [X] Support TLS and HTTP/2. (#20)
- [X] The ability to percent-decode path params. (#21)
- [X] Cookies. (#22)
- [ ] [FlatMap](https://github.com/zacharygolba/via/pull/19#issuecomment-2362172325).
- [ ] Cleanup `via-serve-static`, rename it to `via-fs`, keep `bitflags`.

**Nice to have for the first release:**

- [ ] Proxy middleware as an official crate (`via-proxy`).
- [ ] Multipart support.
- [ ] Tracing.
- [ ] TLS via OpenSSL.
- [ ] Async closure support.

**Things to follow shortly after the first release:**

- [ ] An official crate for compression / decompression.
- [ ] Auth. I'm thinking an official crate with traits and some common adapters (i.e basic, jwt, cookies, etc.).
- [ ] A recipe book or official crate for integrating various templating engines.
- [ ] Compatibility with other async runtimes (tokio is great but being able to chose is also great).
- [ ] Either a basic built-in way of supporting websockets or an official crate with a slightly higher-level API. I don't want to re-invent elixir but it would be nice if we had a story for supporting realtime applications if need be.


crates/via-router/src/iter.rs
```diff
@@ -1,147 +1,247 @@
...
+/// An iterator that yields all possible partial and exact matches for a url path.
 pub struct Visit<'a, 'b, T> {
     node: &'a Node<T>,
-    path: Path<'b>,
-    root: bool,
+    store: &'a Store<T>,
+    depth: usize,
+    index: usize,
+    path_value: &'b str,
+    // TODO: refactor path_segments to be Option<_> and remove the Rc<_> wrapper.
```
zacharygolba, line 32: After running some benchmarks, it turns out that the implementation with `Rc` is faster than the Option wrapper. This is likely due to the value of `path_segments` moving each time the Visit struct is forked.

src/error.rs
```diff
@@ -98,6 +98,40 @@ impl Error {
...
+                if let Some(length) = HeaderValue::from_str(&message.len().to_string()).ok() {
+                    response
+                        .headers_mut()
+                        .insert(http::header::CONTENT_LENGTH, length);
+                }
+
+                *response.status_mut() = status_code;
+                *response.body_mut() = message.into();
+
+                eprintln!("Failed to convert error into response: {}", error);
```
zacharygolba, line 127: I won't leave this eprintln here permanently. I'll refine the code in this part of the match branch, adding comments and finding a better way to indicate why the error couldn't be converted into a response. Maybe an error callback on app is our best option for now?

src/request/path_param.rs
```diff
@@ -1,50 +1,45 @@
 use http::StatusCode;
-use std::{collections::HashMap, str::FromStr};
+use std::str::FromStr;
 
 use crate::{Error, Result};
 
-pub type PathParams = HashMap<&'static str, (usize, usize)>;
+pub(crate) type PathParams = Vec<(&'static str, (usize, usize))>;
```
zacharygolba, line 6: Storing path params in a `Vec` turns out to be faster (and probably safer) than using a `HashMap`. We do the same for query parameters, so maintaining a similar internal API for path params is an added benefit.

src/request/query_param.rs
```diff
@@ -1,13 +1,12 @@
...
 use percent_encoding::percent_decode_str;
-use smallvec::SmallVec;
 use std::{borrow::Cow, slice::Iter, str::FromStr};
 
 use crate::{Error, Result};
 
 pub(crate) type QueryParams<T = String> = Vec<(T, (usize, usize))>;
 
-type ValuesVec<'a> = SmallVec<[&'a (usize, usize); 4]>;
+type ValuesVec<'a> = Vec<&'a (usize, usize)>;
```
zacharygolba, line 9: I initially thought that `SmallVec` would internally fall back to using `Vec` from the standard library. However, this is not the case, and I prefer to use `Vec` since we are storing unbounded user input. I'll remain open to potential alternatives in the future, considering something like the following code snippet if it offers significant performance improvements.

```rust
enum FastVec<T, const N: usize> {
    Stack([Option<T>; N]),
    Heap(Vec<T>),
}
```

crates/via-serve-static/src/resolve.rs
```diff
@@ -6,7 +6,12 @@ use std::{
 };
 use tokio::task;
 
-const CHUNK_SIZE: usize = 1024 * 1024 * 10;
+/// The maximum file size that will be eagerly read into memory. As a performance
+/// optimization, small files will be read into memory immediately to avoid the
+/// overhead of calling `tokio::task::spawn_blocking` multiple times.
+///
+/// The current value is set to `10MB`.
+const EAGER_READ_THRESHOLD: u64 = 1024 * 1024 * 10;
```
zacharygolba, line 14: Since we're not actually streaming `10MB` chunks I decided to change the variable name to something more descriptive. I also thought it was a good idea to change the type def to `u64` to match the return value of `Metadata::len` to prevent us from having to do a type cast for each request.

crates/via-router/src/path.rs
```diff
@@ -13,7 +12,7 @@ pub(crate) enum Pattern {
 /// path separated by `/`.
 pub(crate) struct PathSegments<'a> {
     pub(crate) value: &'a str,
-    segments: SmallVec<[(usize, usize); 6]>,
+    segments: Vec<(usize, usize)>,
```
zacharygolba, line 15: Using `SmallVec` only makes the router between 10 and 20 nanoseconds faster. Considering that the URL path is unbounded user input, I think using the Vec from the standard library provides safety guarantees that make this a reasonable tradeoff.

crates/via-router/src/visitor.rs
```diff
@@ -103,41 +107,52 @@ impl<'a, 'b, T> Visitor<'a, 'b, T> {
...
-                self.matches.push(Match {
-                    path_segment_range,
-                    is_exact_match: self.path.get(depth + 1).is_none(),
-                    node: next,
-                });
-
-                // Continue to match descendants of `next` against the path
-                // segment at the next depth.
-                self.match_at_depth(depth + 1, next);
+            match next.pattern {
```
zacharygolba, line 116: Switching to `match` resulted in the most significant performance improvement. On my machine, the `find_all_matches` benchmark takes around 140 nanoseconds per iteration, down from approximately 170 nanoseconds.

crates/via-router/src/visitor.rs
```diff
@@ -103,41 +107,52 @@ impl<'a, 'b, T> Visitor<'a, 'b, T> {
...
+                        path_segment_range,
+                        is_exact_match,
+                        node: next,
+                    });
+
+                    // Continue to match descendants of `next` against the path
+                    // segment at the next depth.
+                    self.match_at_depth(depth + 1, next);
+                }
+                Pattern::Static(value) if value == path_segment_value => {
```
zacharygolba, line 141: It would be nice to remove the duplication between the match arms of `Pattern::Static` and `Pattern::Dynamic. However, I'm concerned that would result in unnecessary overhead or make the code more difficult to follow.

crates/via-router/src/visitor.rs
```diff
@@ -3,7 +3,32 @@ use crate::{
     routes::{Node, RouteStore},
 };
 
+/// Inserts a match into the `Vec` of matches for a `Visitor` and then
+/// continues to match descendants of `$node` that matched the path segment.
+///
+/// This macro is used to prevent duplication between match branches of nodes
+/// that have a `Pattern::Dynamic` or `Pattern::Static` pattern.
+macro_rules! insert_match_and_continue {
```
zacharygolba, line 11: I don't think duplication is a bad thing, especially since we're calling `match_at_depth` recursively. However, I do believe that preventing drift between the match branches of `Pattern::Dynamic` and `Pattern::Static` is worth the cognitive overhead of a macro. As always, I'm open to suggestions from the community if there is a better way to achieve what I'm trying to achieve here.

zacharygolba:
[575f518](https://github.com/zacharygolba/via/pull/3/commits/575f5182469922e3b09abdac554db7dd278560bb) is my way of saying that I'm sorry. It is a bit verbose, but I think it is a good starting point to ensure that the code in via-router does what it is supposed to do. I'm also hopeful that it inspires the community to further improve the performance of the router.

src/event.rs
```diff
@@ -2,8 +2,7 @@ use std::{net::SocketAddr, sync::Arc};
 
 use crate::Error;
 
-#[derive(Clone, Copy)]
-#[non_exhaustive]
```
zacharygolba: I don't think we'll be introducing any more event variants than necessary. Introducing the event listener callback works well for handling errors that occur outside the request/response cycle and also for the "server ready" callback. However, I believe it's best to use this pattern sparingly in both the private and public API of the framework. I'm open to suggestions if anyone thinks this might result in avoidable breaking changes to the public API in the future.

src/response/body.rs
```diff
@@ -36,17 +37,26 @@ impl Body {
...
 
     pub(super) fn stream(body: BoxStream) -> Self {
         Self {
-            data: Box::pin(StreamBody::new(body)),
+            data: Box::new(StreamBody::new(body)),
             len: None,
         }
     }
+
+    fn project(self: Pin<&mut Self>) -> Pin<&mut BodyData> {
```
zacharygolba, line 52: A bit of scope creep here but it seemed like a good idea to do a proper pin projection for `data` rather than pinning `data` ahead of time and then using `get_mut().data.as_mut()`.

src/response/builder.rs
```diff
@@ -45,6 +45,26 @@ impl ResponseBuilder {
         }
     }
 
+    pub fn headers<I, K, V>(self, headers: I) -> Self
```
zacharygolba, line 48: This will be appreciated by anyone trying to conditionally add a header to a response. I should have included this functionality a long time ago.

crates/via-serve-static/src/stream_file.rs
```diff
@@ -117,53 +116,22 @@ fn stream_file_blocking(sender: mpsc::Sender<ReadChunkResult>, path: PathBuf, ti
...
-                Poll::Pending => return Poll::Pending,
-                // No more data.
-                Poll::Ready(0) => return Poll::Ready(None),
-                // Extend `self.results` with the elements in `buf`.
-                Poll::Ready(_) => {
-                    this.results.extend(buf.into_iter().rev());
-                }
-            }
-        }
+        this.receiver
```
zacharygolba, line 133: The performance of the implementation using `poll_recv_many` is roughly the same as that using `poll_recv`. Given that the `ReadChunkResult` messages are already buffered behind a semaphore in the channel, the Stream implementation using `poll_recv` is likely more secure and correct.

src/middleware/next.rs
```diff
@@ -4,15 +4,18 @@ use super::ArcMiddleware;
 use crate::{BoxFuture, Request, Response, Result};
 
 pub struct Next<State = ()> {
-    stack: VecDeque<ArcMiddleware<State>>,
+    #[allow(clippy::box_collection)]
+    stack: Box<VecDeque<ArcMiddleware<State>>>,
```
zacharygolba, line 8: While boxing a collection can be redundant, I suspect that reducing the size of `Next` from `32` to `8` will result in better overall performance. My reasoning is that it's more important for `Next` to efficiently move between middleware rather than attempting to reduce the indirection introduced by dereferencing the pointer in `Next::call`.

crates/via-serve-static/src/resolve.rs
zacharygolba: `eager_read_threshold` is a configurable option now. I also changed the default value to `1MB`.

src/response/body.rs
```diff
@@ -56,7 +56,7 @@ impl Body {
     }
 
     fn project(self: Pin<&mut Self>) -> BodyProject {
-        // SAFETY:
+        // Safety:
```
zacharygolba, line 59: Additional scope creep here but I want to keep the capitalization consistent.

crates/via-serve-static/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+mod static_file;
+mod stream_file;
+
+use bitflags::bitflags;
+use std::{path::Path, sync::Arc};
+use via::{Endpoint, Error, Result};
+
+use crate::respond::{respond_to_get_request, respond_to_head_request};
+
+pub struct ServeStatic<'a, State> {
```
zacharygolba, line 12: I think I'm going to work on providing some mechanism to have either an owned or borrowed version of middleware. This would allow us to essentially write classes in cases where it make sense. In other words, allow users of the public api to write `async fn call(self: Arc<Self>, request: Request, next: Next) -> Result<Response, Error>`. This would give users of the framework to have immutable state local to middleware which would be incredibly useful for keeping implementation details of community packages from leaking into public apis. Still, I want to support middleware that is called by ref to allow for potentially "happy path"  optimizations. I'm thinking along the lines of a `trait IntoMiddleware` and an enum similar to `std::borrow::Cow` that can hold an owned or by ref Middleware. I'm pretty confident that we're able to pull this off without adding too much complexity the public API (i.e a backwards compatible change). If my thinking is correct, we'll be able to re-write the `ServeStatic` struct to clone `Arc<Self>` and contain the `config` struct along with a semaphore that can either be configured or adapt to `ulimit` to fix the bug when files are streamed to client.

src/app/service.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn project(
+        self: Pin<&mut Self>,
+    ) -> (
+        Pin<&EventListener>,
+        Pin<&mut dyn Future<Output = Result<Response>>>,
+    ) {
+        // Safety:
+        // This block is necessary because we need to project the fields of
+        // `FutureResponse` through the pinned reference.
+        unsafe {
```
zacharygolba, line 93: This is getting fixed tomorrow.

zacharygolba: #11 refactors the app module to get rid unnecessary uses of `unsafe`.

zacharygolba:
Considering the progress made in #11 and #9, here is short summary of what is preventing this PR from merging and the first version of Via from being released.

- [] Refactor the `Request` type and `Error` struct to prevent the need to pass around a channel or `EventListener`.
- [] Use a semaphore or some other mechanism to prevent the server from crashing if there are too many open file descriptors
- [] Support graceful shutdown.
- [] Support https and http2.
- [] Cookies.

Things to follow shortly after the first release:

- [] Tracing.
- [] Multipart support.
- [] An official crate for compression / decompression.
- [] The ability to `flat_map` a request and response body (may be required for compression).
- [] A recipe book or official crate for integrating various templating engines.
- [] Either a basic built-in way of supporting websockets or an official crate with a slightly higher-level API. I don't want to re-invent elixir but it would be nice if we had a story for supporting realtime applications if need be.

src/middleware/middleware.rs
```diff
@@ -0,0 +1,23 @@
...
+pub(crate) type ArcMiddleware<State> = Arc<dyn Middleware<State>>;
+pub type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send + 'static>>;
+
+pub trait Middleware<State>: Send + Sync + 'static {
+    fn call(&self, request: Request<State>, next: Next<State>) -> BoxFuture<Result<Response>>;
+}
+
+impl<State, T, F> Middleware<State> for T
+where
+    T: Fn(Request<State>, Next<State>) -> F + Send + Sync + 'static,
```
zacharygolba, line 15: I'm contemplating putting the bound `Copy` on `T` to prevent middleware functions from capturing their environment. Considering how complex the route definitions can become for an application, it seems like a good idea to require users to write a `struct` with a `Middleware` impl so the intent is clear.

zacharygolba:
If you want to try out Via and want to know what branch you should specify in your Cargo.toml, use `refactor-app-service` for now.

zacharygolba:
I'm considering refactoring the error module and replacing the custom type with an extension trait that works seamlessly with `std::error::Error` and also `serde::Serialize` if the `json` feature flag is enabled.

crates/via-serve-static/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    eager_read_threshold: u64,
+    read_stream_timeout: u64,
+    path_param: &'static str,
+    public_dir: Arc<Path>,
+    flags: Flags,
+}
+
+bitflags! {
+    #[derive(Debug, Clone, Copy)]
+    pub(crate) struct Flags: u8 {
```
zacharygolba, line 30: I like the way we handle config options with `bitflags`. Actix does the same and I admire their static server. We'll definitely keep this in the rewrite.

src/server/io_stream.rs
```diff
@@ -0,0 +1,115 @@
+use hyper::rt::{Read, ReadBufCursor, Write};
+use std::io::{self, IoSlice};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+
+/// A wrapper around a stream that implements both `AsyncRead` and `AsyncWrite`.
+pub struct IoStream<T> {
+    stream: T,
```
zacharygolba, line 9: I think I may want to play around with this a bit before starting the feature work mentioned in the message of this pull request. I want to see how things perform if `stream` is wrapped in an `RwLock`. I haven't tried yet but I imagine benchmarking the echo server with println messages in `poll_read` and `poll_write` may tell a story. I also am considering boxing the stream.

crates/via-router/src/routes.rs
```diff
@@ -0,0 +1,162 @@
...
+                // time you like, but you can never leaf.
+                //
+                // If you see this error, you've probably found a bug in the way
+                // that routes are added to the route store or the way the key
+                // is assigned to nodes.
+                if cfg!(debug_assertions) {
+                    panic!("unknown key: {}", $at);
+                }
+
+                panic!("unknown key");
```
zacharygolba, line 107: I was considering coming up with a non-panicking version of this business logic with hooks to tracing if necessary. However, considering that the only case where this would occur is if the route store's memory is corrupt. In the case of corrupt memory, the safest thing to do is to panic. While this is counterintuitive to the idea of maximum uptime and high availability, it's a choice we make to prefer downtime over sending the wrong or incomplete message. As always, I'm open to feedback as to whether or not my thinking is sound here.

src/app.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::server::serve;
+use crate::{Error, Middleware};
+
+/// The default value of the maximum number of concurrent connections.
+const DEFAULT_MAX_CONNECTIONS: usize = 256;
+
+/// The default value of the shutdown timeout in seconds.
+const DEFAULT_SHUTDOWN_TIMEOUT: u32 = 30;
+
+pub struct App<State> {
```
zacharygolba, line 16: I'm someone with a preference for single letter generic type parameters. However, I noticed while working on this project how beneficial it is when opening a file at a glance, being able to recognize that a type or function takes `State` as a generic parameter. It makes it very easy to how state moves throughout an application with no ambiguity. Sustainability is important to me as I want to be able to dedicate some of time to this long term. For that reason, I'll consider this first and perhaps only exception to my preference of single letter generic type parameters.

zacharygolba:
The FlatMap task could be a Fold. However, I like the idea of FlatMap since the input and output type would always be `Bytes`. We could borrow some of the logic in the implementation of `Buffered` to ensure that whatever is returned from the map fn will be flattened to 8KB chunks. I do like the idea of supporting local state for the map function similar to the accumulator argument passed to fold. In my opinion, that would allow the vast majority of use-cases to benefit from a statically-dispatched body. For all other use-cases, the `http-body-util` crate can be used in combination with `AnyBody::Box(BoxBody)`. Once implemented, the AnyBody enum can be marked as `non_exhuastive`.

PR #8, opened 2024-08-06
feat: bring back the IntoResponse trait

zacharygolba:
Brings back the `IntoResponse` trait. Not having a trait for the conversion wouldn't feel like a Rust framework.


zacharygolba:
This is a good start. I like the idea of limiting the amount of blanket impls for `IntoResponse` to allow the community to define their own.

PR #9, opened 2024-08-11
refactor: prepare req/res body modules for extensibility

zacharygolba:
A work in progress PR that refactors request and response body types to share code when necessary. This refactor also includes a directory structure that should allow for extensibility in the near and long-term future. This includes discrete stream types that either include or filter trailers from request bodies. This implies a potential name change from `.into_stream` to `.into_data_stream` and `.into_stream` where the latter includes trailers.


src/body/response/body.rs
```diff
@@ -37,83 +38,84 @@ enum BodyKind {
...
         }
     }
 
+    // TODO:
+    // Determine if there is a way we can compose closures of nested map
+    // functions to avoid recursive pointers as the value of `BodyKind::Stream`.
+    //
+    // Perhaps we can use a `Box<dyn Fn(Bytes) -> Result<Bytes, E> + Send>` to
+    // store the map functions and unwind them in the `poll_frame` method.
+    pub(crate) fn map<F, E: 'static>(self, _: F) -> Self
```
zacharygolba, line 95: I think the TODO comment is a cool idea but would like some feedback with regards to security and performance with this approach. It seems a bit novel but I like the idea of making a slight compromise here if it makes the code easier to reason about / work with. I also have a feeling the performance may be similar to a statically dispatched version that uses recursive concrete types.

src/body/response/body.rs
```diff
@@ -0,0 +1,225 @@
...
+        T: Stream<Item = Result<Bytes>> + Send + 'static,
+    {
+        let stream = Streaming::new(Box::new(stream));
+
+        Self {
+            body: Either::Left(Either::Right(stream)),
+        }
+    }
+
+    pub(crate) fn map<F>(self, map: Box<F>) -> Self
```
zacharygolba, line 71: Here's a revised version of your PR comment with improved grammar:

---

I think we may want to create a facade struct that contains the `body` field of `ResponseBody` but does not implement the `body` trait. This struct could later be wrapped in another struct that does implement `Body`. This approach would ensure that it is impossible to move data out of the `Pin<&mut Self>` in the `poll_frame` method. While it's not as straightforward as using `map` to `&mut self`, it would give me peace of mind knowing that a mistake that could otherwise go unnoticed would result in a compiler error. Perhaps we can construct the pollable wrapper around `ResponseBody` by mapping the `inner` response before it is returned from `Response::into_inner`.

zacharygolba: I think my concerns have been addressed by [2e9575c](https://github.com/zacharygolba/via/pull/9/commits/2e9575cba112765e3eb91f10858189d9f1f595cf). I'm not going to resolve this conversation in the event that anyone has thoughts or feedback about this approach.

PR #11, opened 2024-08-15
refactor: rework app service with idiomatic use of move

zacharygolba:
A work in progress PR that reworks `App` and `AppService` to depict a more accurate representation of when data and pointers are moved.

---

This pull request grew into something larger than what was originally intended. The changes made here not only improve the performance and stability of Via but also adds necessary features. Considering that the remaining items in the todo list in #3 are features, I'm going to go ahead and rebase and merge this so the remaining items can be worked on incrementally and discretely. I'll keep the branch available until the first release of Via is published on crates.io in case anyone is using a git dependency that references this branch.


src/event.rs
```diff
@@ -21,19 +25,39 @@ pub enum Event<'a> {
...
+        // mitigates the risk of an attacker exploiting `callback`, we'll avoid
+        // having to clone Arc<EventListener> twice for every request.
+        //
+        // Alternatives:
+        // - Use Arc<dyn Fn(Event, &Arc<State>) + Copy + Send + Sync + 'static>
+        // - Use a channel to send events to a stream that exists for the
+        //   lifetime of the application and spawn the event loop that accepts
+        //   incoming TCP connections in a separate tokio task.
+        //
+        (self.callback)(event, state)
```
zacharygolba, line 51: I'm considering refactoring the `App::listen` to return a stream of events as a safer approach to passing around a function pointer and using an `mpsc::channel`. I'm less concerned about the performance of an `Arc::clone` or cloning a `sender` as much as I'm concerned about giving users a safe and idiomatic way of responding the events that occur outside the context of a request / response cycle. While the errors that are bubbled up to the event listener are not critical application errors, knowing of their existence is a game changer when you are tasked with fixing strange or unexpected behavior in an application.

src/server/backoff.rs
```diff
@@ -0,0 +1,33 @@
+use std::time::Duration;
```
zacharygolba, line 1: I plan on adding more comments to this file.

src/server/serve.rs
```diff
@@ -0,0 +1,132 @@
...
+) -> Result<(), Error>
+where
+    State: Send + Sync + 'static,
+{
+    // Create a new backoff strategy with a base delay of 20 milliseconds and a
+    // maximum delay of 30 seconds.
+    let mut backoff = Backoff::new(20, 30);
+    // Create a vector to store the join handles of the spawned tasks. We'll
+    // periodically check if any of the tasks have finished and remove them.
+    let mut handles = Vec::new();
```
zacharygolba, line 39: Keeping a reference to join handles until the task has finished could be a useful metric for tracing in the near future.

docs/examples/hello-world/src/main.rs
```diff
@@ -1,16 +1,11 @@
 use via::{Next, Request, Result};
 
-async fn hello(mut request: Request, _: Next) -> Result<String> {
-    // Attempt to parse the query parameter `n` to a `usize`. If the query
-    // parameter doesn't exist or can't be converted to a `usize`, default to 1.
-    let n = request.query("n").first().parse().unwrap_or(1);
```
zacharygolba: This was a fun way to test out the params / query params API. I didn't mean to commit it.

src/app.rs
```diff
@@ -6,7 +6,7 @@ use crate::router::{Endpoint, Router};
 use crate::server::serve;
 use crate::{Error, Middleware};
 
-const DEFAULT_MAX_CONNECTIONS: usize = 256;
+const DEFAULT_MAX_CONNECTIONS: usize = 248;
```
zacharygolba, line 9: Lowering this value seems to stabilize memory usage when the server experiences high traffic. We should consider providing guidance on setting this value, as there is likely a sweet spot that varies based on runtime, operating system, and use case.

src/request/request.rs
```diff
@@ -73,6 +73,28 @@ impl<State> Request<State> {
...
+    ///     let name = request.param("name").required()?;
+    ///
+    ///     // Create a greeting message with the provided `name`.
+    ///     let message = format!("Hello, {}!\n", name);
+    ///
+    ///     // Send a response with our greeting message, repeated `n` times.
+    ///     Ok(message.repeat(n))
+    /// }
+    /// ```
     pub fn query<'a>(&self, name: &'a str) -> QueryParam<'_, 'a> {
```
zacharygolba, line 98: I like the way this ended up. It may seem a bit silly to wrap `Option` and `Result` the way that we do but it feels really nice to use. Being able to validate the presence of, parse, and then limit a numerical value in query params on 1 line without having to write a utility function that only used 1 or 2 times is a great developer experience. Here is an example of how I modified the hello function in the hello-world example to test everything out.

```rust
use via::{Next, Request, Result};

async fn hello(request: Request, _: Next) -> Result<String> {
    // Attempt to parse the query parameter `n` to a `usize` no greater than
    // 1000. If the query parameter doesn't exist or can't be converted to a
    // `usize`, default to 1.
    let n = request.query("n").first().parse().unwrap_or(1).min(1000);

    // Get a reference to the path parameter `name` from the request uri.
    let name = request.param("name").required()?;

    // Create a greeting message with the provided name.
    let message = format!("Hello, {}!\n", name);

    // Print a message to the console for each friend of the person with `name`.
    for friend in request.query("friend") {
        println!(
            "Hello, {}! Any friend of {} is a friend of mine.",
            name, friend
        );
    }

    // Send a response with our greeting message, repeated `n` times.
    Ok(message.repeat(n))
}
```

src/request/params/query_parser.rs
```diff
@@ -1,103 +1,81 @@
...
-    let mut query_params = Vec::with_capacity(16);
-    let mut input = QueryParserInput::new(input);
-
-    while let Some(query_param) = parse_query_param(&mut input) {
-        query_params.push(query_param);
-    }
+use std::borrow::Cow;
+use std::iter::Peekable;
+use std::str::CharIndices;
+
```
zacharygolba, line 5: This is a bit of a sloppy implementation of a parser but it's functional and sufficient for the weekend. I'll fix this when I come back to work on Monday.

Cargo.toml
```diff
@@ -5,34 +5,23 @@ authors = ["Zachary Golba <zachary.golba@postlight.com>"]
...
-serde = ["dep:serde", "dep:serde_json"]
+json = ["dep:serde", "dep:serde_json"]
 
 [dependencies]
 bytes = "1.7"
 futures-core = { version = "0.3", default-features = false }
 http = "1"
+hyper = { features = ["http1", "server"], version = "1.4" }
 hyper-util = { version = "0.1", features = ["tokio"] }
 percent-encoding = "2.3"
```
zacharygolba, line 18: I'm thinking that once we implement cookies, we'll probably want to feature flag `percent-encoding` and conditionally decode query params based on whether or not that feature flag is set.

zacharygolba: FWIW, I'm thinking `percent-encoding` will be a default feature since It's a pretty common use-case.

src/request/params/query_param.rs
```diff
@@ -0,0 +1,75 @@
+use std::borrow::Cow;
+
+use super::{query_parser::QueryParser, DecodeParam, Param, PercentDecoder};
+
+pub struct QueryParam<'a, 'b> {
```
zacharygolba, line 5: I'm considering making `QueryParam` generic of `DecodeParm` similar to `Param` and then providing an `encoded` or `decode` method that moves self into a `QueryParam<PercentDecode>`. This would give users the choice which could be beneficial if percent-encoded query parameters are not expected. It would also allow us to add a `PercentDecodeLossy` decoder to give users a Via the choice.

src/server/serve.rs
```diff
@@ -57,75 +59,62 @@ where
...
 
-            task::spawn(serve_connection(
-                router,
-                state,
-                permit,
-                io,
-                response_timeout,
-            ))
-        });
+            service_fn(move |request| {
```
zacharygolba, line 77: I plan on writing a struct that impl's Service along with a custom FutureResponse type. I have a feeling it might clean up the code a bit and could even improve performance if the service can be passed to `serve_connection` as `&service`.

zacharygolba: I might keep the `service_fn` depending on how I feel about the implementation.

PR #12, opened 2024-08-19
refactor(via-router): simplify what is passed around in visit

zacharygolba:
A rough draft that attempts to improve the memory usage and performance of via-router by removing unnecessary indirection and favoring references whenever possible.


zacharygolba:
Closing this PR as the work is redundant with what was introduced in #13.

PR #14, opened 2024-09-01
feat(via): wait for connections to close before exit

zacharygolba:
Implements graceful shutdown for the server when a signal is sent to the main process with `ctrl+c`. If connections remain open for longer than the `shutdown_timeout`, an error will be returned from `app.listen` resulting in a non-zero exit code.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,5 +1,6 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::net::{TcpListener, ToSocketAddrs};
 
 use crate::router::{Endpoint, Router};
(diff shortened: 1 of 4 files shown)


src/app.rs
```diff
@@ -10,12 +9,8 @@ use crate::{Error, Middleware};
 /// The default value of the maximum number of concurrent connections.
 const DEFAULT_MAX_CONNECTIONS: usize = 256;
 
-/// The default value of the response timeout in seconds.
-const DEFAULT_RESPONSE_TIMEOUT: u64 = 60;
-
 pub struct App<State> {
     max_connections: usize,
-    response_timeout: Duration,
```
zacharygolba, line 18: I plan on reintroducing this functionality as a built-in middleware exported from the middleware mod.

PR #16, opened 2024-09-06
feat(via): add support for custom body types with Boxed

zacharygolba:
I messed up a rebase which is pretty embarrassing considering that I've been working alone. This PR introduces support for custom response body types.


zacharygolba:
I went ahead and cherry-picked the relevant commits from this branch and opened #19 so we can see a clean diff. This was the easiest way to recover from the rebase gone wrong since the branches didn't diverge that much. Therefore, I'm going to close this in favor of #19. I'll keep the branch up until the end of the day on Monday.

PR #19, opened 2024-09-08
feat(via): add support for custom request and response bodies

zacharygolba:
Adds support for custom response bodies, cleans up some of the code in the `body` and `response` module, and json is supported by default without feature flags.

This is the cleaned up commit history of #16.


src/response/response.rs
```diff
@@ -51,9 +51,14 @@ impl Response {
             .headers(Some(CONTENT_LENGTH).zip(len))
     }
 
+    /// Build a response from a stream of `Result<Frame<Bytes>, Error>`.
+    ///
+    /// If you want to use a stream that is `!Unpin`, you can use the [`Boxed`]
+    /// body in combination with a [`ResponseBuilder`].
+    ///
     pub fn stream<S, E>(stream: S) -> ResponseBuilder
```
zacharygolba, line 59: This is an optimized path for the common case where a `Stream` is `Unpin`. We don't take any functionality away from users by imposing this requirement. You can still use an `!Unpin` stream with a `Boxed` body.

```rust
let stream = NotUnpinStream::new();
let body = Boxed::new(Box::pin(stream));

Response::build().body(body).finish()
```

src/body/boxed.rs
```diff
@@ -11,13 +11,20 @@ pub struct Boxed {
     body: Pin<Box<dyn Body<Data = Bytes, Error = Error> + Send>>,
 }
 
+/// Maps the error type of a body to [Error].
+#[must_use = "streams do nothing unless polled"]
+struct MapError<B> {
```
zacharygolba, line 16: While this adds a small amount of overhead to a Boxed body, it is something that the caller would have to do anyways. An alternative would be using `Box<dyn StdError + Send + Sync>` that may be worth exploring before the first release. Until then, this sufficient and allows for any type that implements `Body<Data = Bytes>` to be used. Here is an updated version of the hello-world example that uses a String in combination with Boxed for the response body.

```rust
use via::body::Boxed;
use via::{Next, Request, Response, Result};

async fn hello(request: Request, _: Next) -> Result<Response> {
    // Get a reference to the path parameter `name` from the request uri.
    let name = request.param("name").required()?;

    Response::build()
        .body(Boxed::new(format!("Hello, {}!", name)))
        .finish()
}
```

src/request/request.rs
```diff
@@ -97,46 +113,43 @@ impl<State> Request<State> {
...
 
-    /// Consumes the request and returns a tuple containing the boxed component
-    /// parts of the request, the body, and an `Arc` to the shared application
-    /// state.
-    pub fn into_parts(self) -> (Parts, RequestBody, Arc<State>) {
-        (self.data.parts, self.data.body, self.data.state)
+    /// Consumes the request and returns a tuple containing the component
+    /// parts of the request and the request body.
+    pub fn into_parts(self) -> (Parts, AnyBody<Box<Incoming>>) {
+        let meta = *self.meta;
```
zacharygolba, line 151: This is safe because we take ownership of request and `self.meta` is immutable and does not expose any `Rc` or `Arc` pointers.

src/request/request.rs
```diff
@@ -172,10 +175,10 @@ impl<State> Debug for Request<State> {
         f.debug_struct("Request")
             .field("method", self.method())
             .field("uri", self.uri())
-            .field("params", &self.data.params)
+            .field("params", &self.meta.params)
             .field("version", &self.version())
             .field("headers", self.headers())
-            .field("body", &self.data.body)
+            // .field("body", &self.data.body)
```
zacharygolba, line 181: I'll fix this before merging.

src/body/pinned.rs
```diff
@@ -21,7 +21,7 @@ impl Pinned {
     /// This method is marked as unsafe because the caller must ensure that the
     /// provided body is `!Unpin` and properly remains pinned in memory while the
     /// body is polled.
-    pub unsafe fn new_unchecked<B>(body: Box<B>) -> Self
```
zacharygolba: I decided not to make this `unsafe` since the caller presumably wrote a pin projection using `unsafe` in order to create the `!Unpin` body. I'm curious to hear how the community feels about this.

src/server/serve.rs
```diff
@@ -44,6 +53,10 @@ where
 
     loop {
         tokio::select! {
+            // A connection being accepted is more likely to occur than the
+            // server shutting down.
+            biased;
```
zacharygolba, line 58: I was originally thinking that it made sense to poll the future returned from the accept function (and also the connection) more frequently than `shutdown_rx.changed()`. However, I think the server benefits from having another future to poll. It creates natural backpressure and it's lock-free. I'm going to remove this.

zacharygolba:
I was planning on adding support for trailers in this pull request. However, it appears that we need to use http2 in order for hyper to write trailer frames at the socket level.

The API of the various body types are already compatible with streams and bodies that yield trailer frames. I confirmed in a simple test that enables the http2 flag and builds the connection in `via::server::serve` with the `http2::Builder` without TLS that the trailers are in fact present in response.

Considering that safety and security is the top priority for this project, I'm thinking that we may want to add HTTP/2 and TLS support (at first via rust-tls) to the list of items blocking the first release in #3. Not having a story for end-to-end encryption definitely feels like it would be a deal breaker for any user of Via with security in mind anyways.

As always, our goal is to build secure and efficient servers with the health care industry, finance industry, and freedom fighters in mind. People need a way to securely share private or sensitive data. This necessity is far more important than the minority of individuals looking for a secure server to hide something that they shouldn't be doing. There will always be bad people in the world that take advantage of software built for the aforementioned purposes. We do not cater to them. We don't build software for them. They are not our users. For that reason, I don't mind if we have to de-prioritize `via-serve-static` out of the first release in order to service our target demographic.

zacharygolba:
Following up on my comment from yesterday. Considering that we allocate 2 file descriptors per connection, I think via-serve-static can be removed from the workspace. Instead, we can provide users of Via with a reasonably fast implementation of IntoResponse for a `StaticFile` and `StreamFile` struct. This allows users to have a simple way of adding a favicon to their API without adding an extra dependency. In the future, for more complex use-cases we can consider official integrations for various cloud storage providers (i.e Google, Amazon, Dropbox, etc.).

zacharygolba:
On second thought, I think a cleaned up version of `via-serve-static` may be worth the effort. The changes I'm thinking of mostly relate to keeping file descriptors open only for as long as we need them (i.e open and close the fd to read metadata rather than hanging on to the fd). We also can use the file size to determine the exact amount of calls to spawn_blocking that we need to optimize file streaming. The implementation can be done using channels or using the return value of the JoinHandle future. Like many decisions that have been made on this project, the option that balances maintainability with performance without excessively favoring either metric is likely the right choice.

zacharygolba:
I think this is ready to merge whenever. I'll wait until the end of the day tomorrow to see if anyone has any comments.

zacharygolba:
Not a requirement for merge but it would be nice if we provided a solution that allowed users to map a body once without having to use BoxBody. I'm not opposed to something restrictive as a built-in implementation of `FlatMap` with a pin projection that doesn't use any macros. We could have it take a function pointer as an argument so the type can be named, worst case scenario we add a second generic parameter to Request and Response. With that being said. The only change I think this needs to merge is marking `AnyBody` as `#[non_exhaustive]`.

PR #20, opened 2024-09-16
feat(via): TLS and HTTP/2 support via rustls

zacharygolba:
Adds support for HTTP/2 and HTTPS via Rustls.


zacharygolba:
This should be working although I haven't had a chance to test the TLS integration yet. I'll add some comments and an example after lunch.

zacharygolba:
Trailers should be working if the http2 feature is enabled. You can test it without using by using the following curl args: `--http2-prior-knowledge -v --raw`.

zacharygolba:
I think I'm going to shift my attention to IoStream as I think there may be some things that I'll need to fix before finishing the rest of this.

docs/examples/tls-rustls/src/main.rs
```diff
@@ -0,0 +1,76 @@
+use rustls::pki_types::{CertificateDer, PrivateKeyDer};
```
zacharygolba, line 1: A lot of this code in this example is ported from https://github.com/rustls/tokio-rustls/blob/main/examples/server.rs. I'll give them credit as a source of inspiration before this PR gets merged.

zacharygolba:
This can merge after #19 unless anyone has any suggestions.

zacharygolba:
I think this can merge whenever. I'll probably wait until the end of the day today (roughly 5 PM UTC−05:00).

PR #22, opened 2024-09-24
feat(via): support request and response cookies

zacharygolba:
Adds support for request and response cookies with the [cookie](https://crates.io/crates/cookie) crate.


src/middleware/cookie_parser.rs
```diff
@@ -0,0 +1,159 @@
...
+        // ownership of the request to the next middleware.
+        //
+        let mut merged_cookies = Box::new(request_cookies.clone());
+
+        Box::pin(async {
+            // Call the next middleware to get a response.
+            let mut response = next.call(request).await?;
+
+            if let Some(cookies) = response.cookies().map(CookieJar::iter) {
+                cookies.cloned().for_each(|cookie| {
```
zacharygolba, line 126: We can probably improve the performance of this middleware by using non `'static` cookies. However, when I think of cookies I think of sessions and authentication. These things require talking to external servers and / or databases which is often times the bottleneck anyways. I think this is a great solution in the short and long term.

PR #27, opened 2024-10-01
fix(via): store param names on the heap

zacharygolba:
Store param names on the heap after they make it out of `via-router`. This ensures that we don't have static references lingering around during async operations.


zacharygolba:
Perhaps we could benefit from some type of string arena in the future but for now, this seems like a reasonable approach.

PR #28, opened 2024-10-02
fix(via-router): impl newtype around Box<str> for param names

zacharygolba:
Refactors via-router to avoid storing static references for param names in favor of a newtype around `Box<str>`. This is a security fix that also stabilizes memory usage.


zacharygolba:
I initially implemented this with `Arc<str>` but ultimately decided that param names are short and we're better off copying them.

PR #32, opened 2024-10-02
fix(via-router): use Arc<str> instead of Box<str>

zacharygolba:
A placeholder PR for testing the performance differences between copying a Box<str> and cloning an Arc<str> in the context of path parameters.


zacharygolba:
Closing this since this ended up getting implemented in #33.

PR #46, opened 2024-10-09
refactor(via): prefer http-body-util types

zacharygolba:
An experimental refactor for a rainy day.


zacharygolba:
I'm unsure how I feel about this. The http-body-util crate is amazing and well-designed to the point that it's hard to not want to use it. This PR definitely makes the code easier to maintain and reason about. However, I like the idea of people being able to see how their data moves throughout an application without having to dig through the docs of a peer dependency or reading rather complex macros. There are some improvements here independent of the move to http-body-util. If nothing else, some of the design improvements and compatibility with the http-body-util crate will make it upstream.

PR #47, opened 2024-10-12
refactor(via): remove the result type alias

zacharygolba:
Removes the `Result` alias type.


zacharygolba:
Closing in favor of #48. It's nice to give users of Via the choice as to whether or not they want to use the shorthand syntax.

PR #61, opened 2024-11-07
feat(via): make the number of file descriptors a config option

zacharygolba:
🚧


zacharygolba:
I'm going to close this in favor of a proxy middleware that properly adjusts backpressure with an exponential backoff if the limit is reached. I'll keep the branch around until the middleware is production ready. Hope everyone is having a great week and is excited for the holiday season to begin.

PR #65, opened 2025-01-07
fix(via-router): use str::match_indices in path::split_into

zacharygolba:
Simplifies the logic used to accumulate path segment ranges in via-router. Also refactors via-router to store path segment ranges in a `Vec` rather than a `SmallVec`. This changes brings a significant performance improvement to via-router and also reduces stack usage in `Router::visit`.

### Benchmarks

The benchmarks are apples to apples. The only difference is that I added additional test cases to see how the router performs as the complexity of the url path increases. 

*Before:*

```
test find_matches_simple       ... bench:         118.79 ns/iter (+/- 2.97)
test find_matches_nested_stack ... bench:         240.54 ns/iter (+/- 12.51)
test find_matches_nested_heap  ... bench:         289.47 ns/iter (+/- 12.16)
```

*After:*

```
test find_matches_2            ... bench:         117.23 ns/iter (+/- 3.17)
test find_matches_3            ... bench:         146.60 ns/iter (+/- 2.76)
test find_matches_4            ... bench:         194.86 ns/iter (+/- 2.38)
test find_matches_5            ... bench:         237.18 ns/iter (+/- 9.49)
test find_matches_6            ... bench:         233.41 ns/iter (+/- 6.69)
test find_matches_7            ... bench:         251.43 ns/iter (+/- 7.11)
```

### DIFF
--- a/crates/via-router/src/lib.rs
+++ b/crates/via-router/src/lib.rs
@@ -7,8 +7,6 @@ mod visitor;
 pub use path::{Param, Span};
 pub use visitor::{Found, VisitError};
 
-use smallvec::SmallVec;
-
 use path::Pattern;
 use routes::{Node, RouteEntry};
 
@@ -49,11 +47,10 @@ impl<T> Router<T> {
     }
 
     pub fn visit(&self, path: &str) -> Vec<Result<Found, VisitError>> {
-        let mut segments = SmallVec::new();
-        let nodes = &self.nodes;
+        let mut segments = vec![];
 
-        path::split_into(&mut segments, path);
-        visitor::visit(path, nodes, &segments)
+        path::split(&mut segments, path);
+        visitor::visit(path, &self.nodes, &segments)
     }
 }
 
--- a/crates/via-router/src/lib.rs
+++ b/crates/via-router/src/lib.rs
@@ -47,7 +47,7 @@ impl<T> Router<T> {
     }
 
     pub fn visit(&self, path: &str) -> Vec<Result<Found, VisitError>> {
-        let mut segments = vec![];
+        let mut segments = Vec::with_capacity(8);
 
         path::split(&mut segments, path);
         visitor::visit(path, &self.nodes, &segments)
--- a/crates/via-router/src/lib.rs
+++ b/crates/via-router/src/lib.rs
@@ -104,7 +104,7 @@ impl<T> Default for Router<T> {
     }
 }
 
-impl<'a, T> Endpoint<'a, T> {
+impl<T> Endpoint<'_, T> {
     pub fn at(&mut self, path: &'static str) -> Endpoint<T> {
         let mut segments = path::patterns(path);
         let key = insert(self.router, &mut segments, self.key);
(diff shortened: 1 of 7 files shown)


crates/via-router/src/path/pattern.rs
```diff
@@ -12,33 +13,16 @@ pub enum Pattern {
...
-    ) => {{
-        let segment = $segment;
-        match segment.get(1..) {
-            // The range is out of bounds or produced an empty str.
-            None | Some("") => panic!($($arg)*),
-            // The range is valid.
-            Some(rest) => rest,
-        }
-    }};
+    ident: Arc<str>,
```
zacharygolba, line 16: Didn't mean to this here! I was experimenting with using `Arc` for params again.

crates/via-router/src/lib.rs
```diff
@@ -49,11 +47,14 @@ impl<T> Router<T> {
     }
 
     pub fn visit(&self, path: &str) -> Vec<Result<Found, VisitError>> {
-        let mut segments = SmallVec::new();
+        let mut segments = Vec::with_capacity(8);
+        let mut results = Vec::with_capacity(8);
```
zacharygolba, line 51: The vectors above are used to store path segments and results for matching routes temporarily. The vectors are used for only a brief duration of time. Therefore, I don't think we have to be concerned about requesting a predictable amount of memory from the heap. I'm curious how the community feels about this.

crates/via-router/src/path/pattern.rs
```diff
@@ -15,30 +15,13 @@ pub struct Param {
...
-            None | Some("") => panic!($($arg)*),
-            // The range is valid.
-            Some(rest) => rest,
-        }
-    }};
-}
-
 /// Returns an iterator that yields a `Pattern` for each segment in the uri path.
 ///
 pub fn patterns(path: &'static str) -> impl Iterator<Item = Pattern> {
```
zacharygolba, line 20: This isn't the most efficient implementation but this isn't a hot code path. This function is not called during a request.

crates/via-router/src/path/span.rs
```diff
@@ -143,9 +90,10 @@ mod tests {
         "/blog/posts/2024//june",
         "/user//profile/settings",
         "/services/contact//us",
+        "/",
```
zacharygolba, line 93: This was a missing edge case that came up during the refactor.

crates/via-router/src/visitor.rs
```diff
@@ -61,16 +60,16 @@ pub struct Found {
 }
 
 pub fn visit(
-    path: &str,
+    results: &mut Vec<Result<Found, VisitError>>,
```
zacharygolba, line 63: I reordered function arguments based on where things should be in memory.

zacharygolba:
The changes in this PR are available via crates.io in version `2.0.0-beta.8`.

crates/via-router/src/path/span.rs
```diff
@@ -8,111 +6,62 @@ pub struct Span {
     end: usize,
 }
```
zacharygolba, line 7: TLDR;

PR #67, opened 2025-01-10
Refactor custom body types

zacharygolba:
A work in progress PR that provides some of our own body types for common use cases while maintaining compatibility with the `http-body-util` crate.


examples/echo-server/src/main.rs
```diff
@@ -1,18 +1,17 @@
+use http::header::CONTENT_TYPE;
 use std::process::ExitCode;
-use via::http::header::CONTENT_TYPE;
-use via::{BoxError, Next, Request, Response, Server};
+use via::{BoxError, Next, Pipe, Request, Response, Server};
```
zacharygolba, line 4: I really like the way this turned out.

zacharygolba:
The changes in this PR are available to test in `v2.0.0-beta.10`.

PR #104, opened 2025-01-27
fix(via): do not set cookie headers in the service_fn

zacharygolba:
Moves the call to `set_cookie_headers` outside of the main service function. This prevents cookies from being set if there isn't a [`cookie_parser`](https://github.com/zacharygolba/via/blob/main/src/middleware/cookie_parser.rs) middleware in your route tree. Be mindful that this is somewhat of a breaking change. The cookie parser middleware must be added before any other middleware functions that use cookies in order for changed cookies to become set cookie headers.

### DIFF
--- a/src/server/io_stream.rs
+++ b/src/server/io_stream.rs
@@ -24,18 +24,15 @@ impl<T> IoStream<T> {
 
 impl<T> IoStream<T> {
     fn project(self: Pin<&mut Self>) -> Pin<&mut T> {
-        // Get a pinned mutable reference to `self.stream` from a pinned mutable
-        // reference to `self`.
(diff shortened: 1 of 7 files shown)


src/server/io_stream.rs
```diff
@@ -24,18 +24,15 @@ impl<T> IoStream<T> {
...
         // Safety:
         //
         // This is safe because data is never moved out of self or
         // `self.stream`.
         //
-        unsafe {
-            let this = self.get_unchecked_mut();
-            Pin::new_unchecked(&mut this.stream)
-        }
+        unsafe { Pin::map_unchecked_mut(self, |this| &mut this.stream) }
```
zacharygolba, line 35: This pin projection is simple enough to be a 1 liner.

PR #123, opened 2025-02-26
fix(via-router): cleanup code in the cache module

zacharygolba:
Cleans up the code in the cache module of via-router for correctness and readability.

### DIFF
--- a/via-router/src/cache.rs
+++ b/via-router/src/cache.rs
@@ -1,24 +1,13 @@
 use std::collections::VecDeque;
 use std::error::Error;
-use std::fmt::{self, Display, Formatter};
 use std::sync::RwLock;
 
-use crate::router::{Match, Matched};
-
-#[derive(Debug)]
-pub struct CacheError;
+use crate::router::Match;
 
 pub struct Cache {
     capacity: usize,
-    entries: RwLock<VecDeque<(Box<str>, Matched)>>,
-}
-
-impl Error for CacheError {}
-
-impl Display for CacheError {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        write!(f, "lock in use")
-    }
+    #[allow(clippy::type_complexity)]
+    entries: RwLock<VecDeque<(Box<str>, Vec<Option<Match>>)>>,
 }
 
 impl Cache {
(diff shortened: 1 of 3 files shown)


via-router/src/cache.rs
```diff
@@ -1,29 +1,24 @@
...
 pub struct Cache {
     capacity: usize,
-
-    #[allow(clippy::type_complexity)]
-    entries: RwLock<VecDeque<(Box<str>, Vec<Option<Match>>)>>,
+    entries: RwLock<VecDeque<(Box<str>, Matched)>>,
 }
 
-#[inline]
-fn fetch(
```
zacharygolba, line 14: To my understanding, there is nothing wrong with calling to_vec on a `&Vec<_>` in find_map since it is only called once if a match is found. That was the only reason for having a separate function here.

via-router/src/cache.rs
```diff
@@ -34,37 +29,33 @@ impl Cache {
...
+    pub fn promote(&self, key: usize) {
+        if let Ok(mut guard) = self.entries.try_write() {
+            if self.capacity.checked_div(2).is_some_and(|half| key > half) {
                 guard.swap_remove_front(key);
             }
         }
+    }
 
-        Some(Some(matches))
+    pub fn read(&self, path: &str) -> Result<Option<(usize, Matched)>, CacheError> {
```
zacharygolba, line 40: Returning an error here is better in my opinion because it signals intent to the caller.

via-router/src/cache.rs
```diff
@@ -34,37 +29,33 @@ impl Cache {
...
+        let guard = self.entries.try_read().or(Err(CacheError))?;
```
zacharygolba, line 41: I typically use if..let for mutex and rw-lock guards but this seems like a reasonable exception.

via-router/src/cache.rs
```diff
@@ -34,37 +29,33 @@ impl Cache {
...
-            return match lock.try_read() {
-                Ok(guard) => match fetch(&guard, path) {
-                    Some(existing) => break existing,
-                    None => Some(None),
-                },
-                Err(_) => None,
-            };
-        };
-
-        if cap.checked_div(2).is_some_and(|half| key > half) {
```
zacharygolba, line 52: This business logic is now the responsibility of the caller with the introduction of `Cache::promote`.

via-router/src/router.rs
```diff
@@ -156,8 +175,11 @@ impl<T> Router<T> {
 
     #[inline]
     pub fn resolve(&self, matching: &Match) -> (Option<&Param>, Option<&T>) {
```
zacharygolba, line 177: I was hungry when I wrote this...

PR #136, opened 2025-02-28
fix(via): remove the nested accept loop in serve

zacharygolba:
Favors simplicity by removing the nested accept loop in the serve function under the assumption that it will likely result in less complicated bitcode once compiled.

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -66,9 +66,38 @@ where
     });
 
     // Start accepting incoming connections.
-    let exit_code = 'accept: loop {
+    let exit_code = loop {
         // Acquire a permit from the semaphore.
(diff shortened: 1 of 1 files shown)


src/server/serve.rs
```diff
@@ -207,8 +207,7 @@ where
 
         // Otherwise, return an error.
         _ = time::sleep(shutdown_timeout) => {
-            let message = "server exited before all connections were closed".to_owned();
-            Err(DynError::from(message))
+            Err("server exited before all connections were closed".into())
```
zacharygolba, line 210: Slight scope creep here but it's not a hot path, simple enough of a change, and doesn't warrant an additional commit message.

PR #140, opened 2025-03-03
feat(via): basic file system support

### LINKED ISSUE #63: Impl IntoResponse for File
Refactoring the static-server middleware can come in 2 phases. First, we can implement IntoResponse for File in via proper. In the future we can decide whether or not to implement an official static server middleware. I like the idea of seeing what the community comes up with.

The criteria for closing this issue is as follows. Create an impl IntoResponse for std::fs::File and tokio::fs::File (both requiring a feature flag). The implementation should respond with the entire contents of the File buffered in memory and should error if the File is larger than some constant limit imposed at the callsite with builder pattern similar to actix.

A streaming file implementation can be added either as a part of one builder struct or two discrete types. It can also be done in a separate pull request with a separate issue. I'm open to bikeshedding about the best approach here if anyone wants to chime in.

zacharygolba:
Closes #63.

---

Description coming tomorrow along with docs and maybe even a test. I'm also going to make sure that a single file descriptor is used for both the metadata and reading the data. I'll probably rename a couple of things as well. I want to keep the name `File` for the builder struct though. We're about simplicity and I would like that to be obvious.

I'll follow up with an implementation to stream files but it's a low priority since I'm a big fan of using cloud based file serving solutions for serious production apps. If I'm being honest with you all, I'd probably only ever use the favicon middleware in production for the reasons mentioned in the comments below. There are so many great cloud based options available today with features that I'd rather not implement myself.


src/response/file.rs
```diff
@@ -0,0 +1,92 @@
...
+use tokio::fs;
+
+use super::Response;
+use crate::{middleware, Error};
+
+type GenerateEtag = fn(&Metadata) -> Result<String, Error>;
+
+/// Serve a single file from disk.
+///
+pub struct File {
```
zacharygolba, line 15: I'm considering using references for all of these variables but want to see where the await points are before assuming things can be borrowed.

I would really like to use `&'a dyn FnOnce(&Metadata) -> Result<String, Error>` for the generate etag struct field but think that would require an additional `Box` and these seem like a lot of unnecessary allocations. It's important that this is sound from an async perspective but I'm going to try to come up with a better way. Using a function pointer may feel gross but it's likely our best path forward.

zacharygolba: In other words, if an attacker can overwrite the value of the function pointer between await points to a nefarious function of their own creation–we have bigger problems to worry about.

Cargo.toml
```diff
@@ -10,7 +10,9 @@ homepage = "https://github.com/zacharygolba/via"
 repository = "https://github.com/zacharygolba/via"
 
 [features]
-default = ["http1"]
+default = ["fs", "http1"]
```
zacharygolba, line 13: My thinking here is that most servers will at least want to use the fs feature to serve the favicon because it's a best practice to do so.

If you are concerned about denial of service attacks coming from requesting a favicon, you should have a cache in front of your server anyways.

In practice, if the decision was up to me I would either

1. filter traffic requesting a favicon to nodes running the app server at either the API gateway or the load balancer if it's fancy enough. This could introduce a regex based denial of service attack albeit unlikely.

2. Put a cache in front on my API for get requests, make sure to include the favicon path, and set the expiry for the key associated with the favicon to never expire. I would probably prefer to manually invalidate the cache for a favicon because it doesn't change often and there typically aren't many permutations of the keys associated with them.

src/response/file.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub async fn serve(self) -> middleware::Result {
+        let path = self.path.as_path();
+        let data = fs::read(path).await.map_err(wrap_io_error)?;
+        let meta = if self.etag.is_some() || self.with_last_modified {
+            Some(fs::metadata(path).await.map_err(wrap_io_error)?)
+        } else {
+            None
+        };
+
+        let mut response = Response::build().header(CONTENT_LENGTH, data.len());
```
zacharygolba, line 73: I don't mind reassigning to `response` for the conditional parts of the builder. This code is sync for the remainder of the function.

src/response/file.rs
```diff
@@ -0,0 +1,94 @@
...
+            ..self
+        }
+    }
+
+    pub async fn serve(self) -> middleware::Result {
+        let (data, metadata) = {
+            let mut file = fs::File::open(&self.path).await.map_err(wrap_io_error)?;
+            let metadata = file.metadata().await?;
+
+            let mut data = match metadata.len().try_into() {
```
zacharygolba, line 68: This is probably less performant than allocating just in time. However, I prefer to fail early due to an allocation error while the buffer is empty.

src/response/file.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    pub fn with_last_modified(self) -> Self {
+        Self {
+            with_last_modified: true,
+            ..self
+        }
+    }
+
+    pub async fn serve(self) -> middleware::Result {
+        let mut file = fs::File::open(&self.path).await.map_err(wrap_io_error)?;
```
zacharygolba, line 64: This could be refactor to a block or a separate function but I think I prefer to hang on to the underlying fd until the response is generated.

src/middleware/favicon.rs
```diff
@@ -0,0 +1,46 @@
...
+/// #[tokio::main(flavor = "current_thread")]
+/// async fn main() -> Result<(), Error> {
+///     let mut app = via::app(());
+///
+///     app.at("/favicon.ico").respond(favicon("./favicon.ico"));
+///     Ok(())
+/// }
+/// ```
+///
+pub fn favicon<T>(path: impl AsRef<Path>) -> impl Middleware<T> {
```
zacharygolba, line 26: I'm considering removing this middleware in favor of using the following in an example or guide somewhere.

```rust
app.at("/favicon.ico").respond(via::get(|_, _| {
    File::open(PathBuf::from(PUBLIC_DIR).join("favicon.png"))
        .content_type("image/x-icon".to_owned())
        .with_last_modified()
        .serve()
}));
```

Cargo.toml
zacharygolba: I'm considering not making the file server a default feature. While it's important for many servers to have a favicon, this is Rust and perhaps there are enough use cases that are low-level and do not need this feature out of the box. Especially if it's this easy to implement the logic yourself.

```rust
app.at("/favicon.ico").respond(via::get(|_, _| {
    File::open(PathBuf::from(PUBLIC_DIR).join("favicon.png"))
        .content_type("image/x-icon".to_owned())
        .with_last_modified()
        .serve()
}));
```

Cargo.toml
```diff
@@ -23,6 +25,7 @@ futures-core = { version = "0.3", default-features = false }
 http = "1"
 http-body = "1"
 http-body-util = "0.1"
+httpdate = "1"
```
zacharygolba, line 28: I'm probably going to mark this as optional even though it's already a dependency of hyper. The change will do nothing other than get me in the habit of making sure that dependencies that are only used for a feature flag are specified the way that you would expect in Cargo.toml.

Cargo.toml
zacharygolba: Definitely keeping `fs` as a feature flag and not making it a default feature. Will push up the change now and then this can merge.

src/response/file.rs
```diff
@@ -73,15 +64,14 @@ impl File {
...
-        let metadata = file.metadata().await?;
+        let mut file = fs::File::open(&self.path).await.map_err(Error::from_io)?;
+        let metadata = file.metadata().await.map_err(Error::from_io)?;
 
-        let mut data = match metadata.len().try_into() {
-            Ok(capacity) => Vec::with_capacity(capacity),
-            Err(error) => return Err(Error::payload_too_large(error.into())),
-        };
+        // Allocate the exact capacity required to store the file in memory.
+        // This is
```
zacharygolba, line 71: I think I'm going to introduce a stream implementation here if the file is larger than `15MB`. I made the decision mid comment.

zacharygolba:
The streaming file impl can be found here: https://github.com/zacharygolba/via-backup/blob/main/src/response/file.rs#L70
The working example can be found here: https://github.com/zacharygolba/via-backup/tree/main/examples/file-server

Screenshot 2025-03-05 alle 5 36 33 PM

zacharygolba:
Just tried to publish a release with the file feature from with `--allow-dirty` from my terminal and it doesn't look like the changes made it to the release for some reason. cc @wycats

zacharygolba:
Just tried to publish an additional release (obv wouldn't do this on a stable release channel)...

zacharygolba:
Credit to [@von_co](https://unsplash.com/it/@von_co) for the photo in the about example in the commits that I'm unable to push. I love your art! ❤️

zacharygolba:
So strange that running cargo publish doesn't include the new file API or any commit from this branch. I published 2 releases last night, both times I confirmed that I was on the right branch by running the file-server example in an adjacent terminal session.

I also made a fresh clone of the project, checked out this branch, copied the changes over (excluding all dot files), and tried to push the uncommitted changes but received the same error. It seems to be coming from curl.

If anyone has any advice, it would be greatly appreciated. This is a nice feature to have. Especially for benchmarks and examples.

```
> git push
Enumerazione degli oggetti in corso: 56, fatto.
Conteggio degli oggetti in corso: 100% (56/56), fatto.
Compressione delta in corso, uso fino a 10 thread
Compressione oggetti in corso: 100% (35/35), fatto.
Scrittura degli oggetti in corso: 100% (37/37), 1.04 MiB | 42.52 MiB/s, fatto.
37 oggetti totali (28 delta), 0 riutilizzati (0 delta), 0 riutilizzati nel file pack
error: RPC non riuscita; HTTP 400 curl 22 The requested URL returned error: 400
send-pack: unexpected disconnect while reading sideband packet
fatal: l'interlocutore remoto ha chiuso inaspettatamente la connessione
Everything up-to-date
```

I stumbled upon an answer from stack overflow stating that I should increase the git buffer size. I didn't try it because I find it hard to believe that after all the refactoring I've done without issue, a couple of average sized commits to a feature branch require system wide changes to git config.

zacharygolba:
I also haven't had a job since June of last year and have been going through a somewhat difficult time. If it seems like I'm making a big deal about this it's because this project is one of the only things keeping me going. Hope this gets resolved soon.

zacharygolba:
Looks like I'm able to push from GitHub Desktop. Not sure what is going on with the CLI... Will investigate after this lands.

src/body/stream_body.rs
```diff
@@ -36,7 +36,7 @@ impl<T> StreamBody<T> {
 
 impl<T> Body for StreamBody<T>
 where
-    T: Stream<Item = Result<Frame<Bytes>, DynError>> + Send + Sync,
+    T: Stream<Item = Result<Bytes, DynError>> + Send + Sync,
```
zacharygolba, line 39: I think I'm going to impose an `Unpin` requirement here for API safety reasons. If someone needs to use a `!Unpin` stream as the source for a response body, they can do so by using BoxBody directly with `Response::new` constructor method.

zacharygolba:
Heads up, releases `rc.20` and `rc.21` do not contain the code that was present in the working directory when I ran cargo publish. I'm going to yank both versions and push up a release now.

PR #141, opened 2025-03-03
refactor(via): response::Builder ~> ResponseBuilder

zacharygolba:
Renames `response::Builder` to `ResponseBuilder` so the structs in the response module doc stack like this:

Screenshot 2025-03-03 alle 5 58 31 PM


zacharygolba:
This seems silly but I like being explicit here.

PR #145, opened 2025-03-05
feat(via): a way to go from io::Error to Error

papa-zakari:
Provides an explicit way to convert an `io::Error` to an `Error` in the public API.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,6 +7,7 @@ use serde::ser::SerializeStruct;
 use serde::{Serialize, Serializer};
 use std::error::Error as StdError;
 use std::fmt::{self, Debug, Display, Formatter};
+use std::io;
 
 use crate::response::Response;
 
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -34,12 +35,61 @@ struct SerializeError<'a> {
...
     ///
     pub fn new(source: DynError) -> Self {
         Self::internal_server_error(source)
     }
 
+    /// Create a new [`Error`] from the provided [`io::Error`]. The status code
+    /// of the error returned will correspond to `source.kind()`.
+    ///
+    pub fn from_io(source: io::Error) -> Self {
+        match source.kind() {
```
zacharygolba, line 48: There shouldn't be an issue allocating in a match branch. My reason for mentioning it is in case someone knows otherwise.

src/error.rs
```diff
@@ -34,12 +35,61 @@ struct SerializeError<'a> {
...
+            }
+
+            io::ErrorKind::TimedOut => {
+                // Implies an upstream service timeout.
+                Self::gateway_timeout(Box::new(source))
+            }
+
+            _ => {
+                // Any other kind is treated as an internal server error.
+                Self::internal_server_error(Box::new(source))
```
zacharygolba, line 88: I want to be conservative here so that way an opinion of what a status code should be doesn't preclude someone from using this function liberally.

zacharygolba: Also, feel free to submit a PR if you see something that is missing here. I like the idea of highly opinionated parts of the code being evolved over time by heated debates within the community. As long as were all professional, treat each other with respect, and follow the [CODE_OF_CONDUCT.md](https://github.com/zacharygolba/via/blob/main/CODE_OF_CONDUCT.md), I'm game.

PR #150, opened 2025-03-07
refactor(via): rename Server::new ~> via::start

zacharygolba:
Renames `Server::new` to `via::start`. It feels a bit strange using constructor syntax for the server where typically, in Via code–it is reserved for data. This change is inspired both by the themed `.ignite()` function in [Rocket](https://rocket.rs/) as well as the functional API seen in [warp](https://github.com/seanmonstar/warp).


zacharygolba:
Doc changes coming up!

PR #153, opened 2025-03-07
refactor(via): cleanup server code w/ simplicity in mind

zacharygolba:
A work in progress refactor that cleans up the code in the server module with simplicity in mind... I don't anticipate it changing much more than this unless anyone has any suggestions. Yes. I'm asking for help everyone!

### DIFF
--- a/src/server/serve.rs
+++ b/src/server/serve.rs
@@ -20,28 +20,20 @@ use crate::request::Request;
 use crate::router::{Router, RouterError};
 use crate::server::io_stream::IoStream;
 
-pub async fn serve<T, A>(
-    listener: TcpListener,
-    mut acceptor: A,
-    state: T,
-    router: Router<T>, // Router holds no references to T.
+pub async fn serve<T: Send + Sync + 'static>(
+    state: Arc<T>,
+    router: Arc<Router<T>>, // Router holds no references to T.
+    mut acceptor: impl Acceptor + 'static,
     max_connections: usize,
     max_request_size: usize,
     shutdown_timeout: Duration,
-) -> Result<ExitCode, DynError>
-where
-    T: Send + Sync + 'static,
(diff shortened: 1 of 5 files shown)


src/server/serve.rs
```diff
@@ -145,7 +141,7 @@ where
                 let matches = router.visit(request.uri().path());
                 let next = router.resolve(&matches, request.params_mut());
 
-                async {
+                Box::pin(async {
```
zacharygolba, line 144: A bit of scope creep here but probably beneficial if there is an additional layer of indirection here. This is the boundary where library code stops and app code begins.

zacharygolba: This means that the majority of the library code will be work with data on the stack. App code, will have a boxed future for every middleware function that is called + 1. I'm assuming that a `Pin<Box<Future>>` is required for the work stealing thread pool in tokio to poll a future created from a separate task. I'm also assuming that this is what we want in that, the connection will be polled and bytes will be written to the socket from the task spawned when the connection is accepted while app code can be polled from any task (worker thread). Curious if anyone on the tokio team has thoughts or opinions here.

src/server/server.rs
```diff
@@ -173,27 +144,28 @@ impl<T: Send + Sync + 'static> Server<T> {
...
-            address,
-            app,
-            max_connections,
-            max_request_size,
-            shutdown_timeout,
+        // Serve incoming connections until shutdown is requested.
+        serve(
+            Arc::clone(&self.state),
+            Arc::clone(&self.router),
+            self.acceptor()?,
```
zacharygolba, line 159: I like the idea of the `Server` struct being intact until the server exits. It's a best practice with async code. If having one additional reference to `state` or `router` ever becomes a problem, we can reconsider as this isn't a breaking change.

PR #159, opened 2025-03-09
fix(via): potential overflow in FileStream

zacharygolba:
Fixes a potential overflow that could occur if the file size is somehow modified while it is open. While unlikely, it's something I would prefer to make explicitly impossible. Also removes the use of Option in `FileStream`, which improves security by leaving `FileStream` intact to the extent that it (maybe?) could be !Unpin.

### DIFF
--- a/src/response/file.rs
+++ b/src/response/file.rs
@@ -25,7 +25,7 @@ const BASE_DELAY_IN_MILLIS: u64 = 100;
 
 /// The amount of times that we'll retry to open a file if in an error occurs.
 ///
-const MAX_ATTEMPTS: u64 = 3;
+const MAX_ATTEMPTS: u64 = 5;
 
 /// A function pointer used to generate an etag.
 ///
@@ -59,7 +59,7 @@ enum Open {
 struct FileStream {
     remaining: usize,
     buffer: Vec<MaybeUninit<u8>>,
-    file: Option<TokioFile>,
+    file: TokioFile,
 }
 
 /// Attempt to open the file and access the metadata at the provided path.
@@ -210,7 +210,7 @@ impl FileStream {
         Self {
             remaining,
             buffer: vec![MaybeUninit::uninit(); MAX_FRAME_LEN],
-            file: Some(file),
+            file,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/response/file.rs
```diff
@@ -59,7 +59,7 @@ enum Open {
 struct FileStream {
     remaining: usize,
     buffer: Vec<MaybeUninit<u8>>,
-    file: Option<TokioFile>,
+    file: TokioFile,
```
zacharygolba, line 62: Local benchmarks show that FileStream is dropped when expected meaning that Option is unnecessary here. I'm curious how perform is across platforms though. Having to change this is rather insignificant with regards to effort. Therefore, I prefer to error on the side of caution.

src/response/file.rs
```diff
@@ -219,31 +219,31 @@ impl Stream for FileStream {
     type Item = Result<Bytes, DynError>;
 
     fn poll_next(self: Pin<&mut Self>, context: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let this = self.get_mut();
-
-        if this.remaining == 0 {
-            this.file = None;
-            this.buffer.fill(MaybeUninit::uninit());
+        if self.remaining == 0 {
```
zacharygolba, line 222: `remaining` is a Copy type and I prefer to have the preconditions happen immediately.

src/response/file.rs
```diff
@@ -219,31 +219,31 @@ impl Stream for FileStream {
...
+        match Pin::new(&mut this.file).poll_read(context, &mut data) {
             Poll::Pending => Poll::Pending,
 
             Poll::Ready(Ok(())) => {
                 let filled = data.filled();
-                this.remaining -= filled.len();
-                Poll::Ready(Some(Ok(Bytes::copy_from_slice(filled))))
+
+                if let Some(remaining) = this.remaining.checked_sub(filled.len()) {
+                    this.remaining = remaining;
```
zacharygolba, line 236: The code is sync here and not in a loop. It's easier to work directly with `this` as if it was `self` rather than storing `let remaining = &mut this.remaining`... I feel the same way about buffer. If we were in a loop, I'd treat things differently.

src/response/file.rs
```diff
@@ -25,7 +25,7 @@ const BASE_DELAY_IN_MILLIS: u64 = 100;
 
 /// The amount of times that we'll retry to open a file if in an error occurs.
 ///
-const MAX_ATTEMPTS: u64 = 3;
+const MAX_ATTEMPTS: u64 = 5;
```
zacharygolba, line 28: Heads up, I'm slightly biased against not wanting to make the fastest file server implementation in favor of an API that is fun to use, correct, more than fast enough, easy to audit, and easy to contribute to.

I decided to increase attempts here as it takes slightly longer for `FileStream` to get dropped than it did to explicitly set `this.file = None`. We could of course, work with a raw file descriptor but that could become overly complicated.

PR #162, opened 2025-03-09
fix(via): unconditionally uninit bytes in FileStream

zacharygolba:
Unconditionally cleans up the bytes read into the buffer after they are copied to a cloneable bytes instance. Releases will planned and less like a nightly build with time for code review and me to sleep on changes in the future.

### DIFF
--- a/src/response/file.rs
+++ b/src/response/file.rs
@@ -231,19 +231,16 @@ impl Stream for FileStream {
 
             Poll::Ready(Ok(())) => {
                 let filled = data.filled();
+                let len = filled.len();
 
-                if let Some(remaining) = this.remaining.checked_sub(filled.len()) {
+                if let Some(remaining) = this.remaining.checked_sub(len) {
                     let data = Bytes::copy_from_slice(filled);
                     this.remaining = remaining;
-
-                    if remaining == 0 {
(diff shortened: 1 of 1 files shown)


zacharygolba:
Pretty soon I'm going to run out of money for coffee and the cable bill. In a sense, I'm racing against the clock. `v2.0.0-rc.27` coming up.

src/response/file.rs
```diff
@@ -231,19 +231,16 @@ impl Stream for FileStream {
...
-                if let Some(remaining) = this.remaining.checked_sub(filled.len()) {
+                if let Some(remaining) = this.remaining.checked_sub(len) {
                     let data = Bytes::copy_from_slice(filled);
                     this.remaining = remaining;
-
-                    if remaining == 0 {
-                        this.buffer.fill(MaybeUninit::uninit());
-                    }
-
+                    this.buffer[..len].fill_with(MaybeUninit::uninit);
```
zacharygolba, line 239: I recently changed my mind about preferring `(&mut this.bufer[..len])` after thinking about it and realizing this happens anytime that a method that takes `&self` is called from method that takes an owned version of `self`.

Issue #146, opened 2025-03-05
Last minute HttpBody API changes


zacharygolba:
I'm thinking that prior to releasing the official version `2.0` of Via, we should add a `tee()` method to request and response. Perhaps an `Pin<Box<dyn AsyncWrite + Send + Sync>>`. In order to facilitate this, we can modify the `HttpBody` enum to the following and then mark it as `non_exhuastive`. I imagine that this would be useful in industries where I see Via being most important. Such as finance and healthcare where just-in-time low cost auditing is a requirement.

```rust
/// The body of a request or response.
///
#[must_use = "streams do nothing unless polled"]
pub enum HttpBody<T> {
    /// The original body of a request or response.
    ///
    Inline(T),

    /// A type erased, dynamically dispatched [`Body`].
    ///
    Dyn(BoxBody),

    /// A box body that writes a reference to each frame into a
    /// [`dyn AsyncWrite`](AsyncWrite).
    ///
    Tee(BoxBody, Pin<Box<dyn AsyncWrite + Send + Sync>>),
}
```

zacharygolba:
I'm really curious how people feel about `tee` being on the body rather than the socket for the sake of auditing. IMO, it's probably preferable that an `Authorization` header or `Set-Cookie` header doesn't make it's way into paper trail.

zacharygolba:
It's unlikely that we'll implement this feature until given an explicit reason to do so.

#164 can be referenced if we end up implementing `HttpBody::flat_map()` `HttpBody::fold()` in house.

PR #164, opened 2025-03-10
feat(via): Request::tee and Response::tee

zacharygolba:
A work in progress implementation of #146.

---

This should be the last breaking API change before an official 2.0 release. I plan on waiting until we confirm that no public API changes are required for the sake of safety before publishing the release. Tentatively scheduled for an April Fools Day release.

### DIFF
--- a/examples/echo-server/Cargo.toml
+++ b/examples/echo-server/Cargo.toml
@@ -5,5 +5,5 @@ edition = "2021"
 
 [dependencies]
 http = "1"
-tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "io-std"] }
 via = { path = "../.." }
--- a/examples/echo-server/Cargo.toml
+++ b/examples/echo-server/Cargo.toml
@@ -5,5 +5,5 @@ edition = "2021"
 
 [dependencies]
 http = "1"
-tokio = { version = "1", features = ["macros", "rt-multi-thread", "io-std"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "io-util"] }
 via = { path = "../.." }
--- a/examples/echo-server/Cargo.toml
+++ b/examples/echo-server/Cargo.toml
@@ -5,5 +5,5 @@ edition = "2021"
 
 [dependencies]
 http = "1"
-tokio = { version = "1", features = ["macros", "rt-multi-thread", "io-util"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 via = { path = "../.." }
--- a/examples/echo-server/src/main.rs
+++ b/examples/echo-server/src/main.rs
@@ -1,5 +1,6 @@
 use http::header::CONTENT_TYPE;
 use std::process::ExitCode;
+use tokio::io::stderr;
 use via::middleware::error_boundary;
 use via::{Next, Pipe, Request, Response};
 
@@ -23,8 +24,20 @@ async fn main() -> Result<ExitCode, Error> {
     let mut app = via::app(());
 
     app.include(|request: Request, next: Next| {
-        let request = request.tee(tokio::io::stderr());
-        async { Ok(next.call(request).await?.tee(tokio::io::stderr())) }
+        let request = request.map(|body| body.tee(stderr()));
+        //
+        // For additional indirection you could alternatively do...
+        // let request = request.map(|body| body.tee(stderr()).boxed());
+        //
+
+        async {
+            let response = next.call(request).await?;
+            Ok(response.map(|body| body.tee(stderr())))
+            //
+            // For additional indirection you could alternatively do...
+            // Ok(response.map(|body| body.tee(stderr()).boxed()))
+            //
+        }
     });
 
     // Include an error boundary to catch any errors that occur downstream.
--- a/examples/echo-server/src/main.rs
+++ b/examples/echo-server/src/main.rs
@@ -1,6 +1,5 @@
 use http::header::CONTENT_TYPE;
 use std::process::ExitCode;
-use via::body::{BoxBody, TeeBody};
 use via::middleware::error_boundary;
 use via::{Next, Pipe, Request, Response};
 
@@ -28,21 +27,9 @@ async fn main() -> Result<ExitCode, Error> {
         eprintln!("Error: {}", error);
     }));
 
-    // Map the request and response bodies to a boxed body type with 2 levels
-    // of indirection. Also, connect a sink a sync to the request and response
-    // body for the purpose of audit logs (because enterprise software).
     app.include(|request: Request, next: Next| {
-        let request = request.map(|body| {
-            let sink = tokio::io::stderr();
-            BoxBody::new(TeeBody::new(body.boxed(), sink))
-        });
-
-        async {
-            Ok(next.call(request).await?.map(|body| {
-                let sink = tokio::io::stdout();
-                BoxBody::new(TeeBody::new(body.boxed(), sink))
-            }))
-        }
+        let request = request.tee(tokio::io::stderr());
+        async { Ok(next.call(request).await?.tee(tokio::io::stdout())) }
     });
 
     // Add our echo responder to the endpoint /echo.
--- a/examples/echo-server/src/main.rs
+++ b/examples/echo-server/src/main.rs
@@ -32,19 +32,15 @@ async fn main() -> Result<ExitCode, Error> {
     // of indirection. Also, connect a sink a sync to the request and response
     // body for the purpose of audit logs (because enterprise software).
     app.include(|request: Request, next: Next| {
-        let future = next.call(request.map(|existing| {
-            BoxBody::new(TeeBody::new(
-                BoxBody::new(existing.into_inner()),
-                tokio::io::stderr(),
-            ))
-        }));
+        let request = request.map(|body| {
+            let sink = tokio::io::stderr();
+            BoxBody::new(TeeBody::new(body.boxed(), sink))
+        });
 
         async {
-            Ok(future.await?.map(|existing| {
-                BoxBody::new(TeeBody::new(
-                    BoxBody::new(existing.into_inner()),
-                    tokio::io::stdout(),
-                ))
+            Ok(next.call(request).await?.map(|body| {
+                let sink = tokio::io::stdout();
+                BoxBody::new(TeeBody::new(body.boxed(), sink))
             }))
         }
     });
--- a/examples/echo-server/src/main.rs
+++ b/examples/echo-server/src/main.rs
@@ -1,6 +1,6 @@
 use http::header::CONTENT_TYPE;
 use std::process::ExitCode;
-use tokio::io::stderr;
+use tokio::io;
 use via::middleware::error_boundary;
 use via::{Next, Pipe, Request, Response};
 
@@ -24,18 +24,18 @@ async fn main() -> Result<ExitCode, Error> {
     let mut app = via::app(());
 
     app.include(|request: Request, next: Next| {
-        let request = request.map(|body| body.tee(stderr()));
+        let request = request.map(|body| body.tee(io::empty()));
         //
         // For additional indirection you could alternatively do...
-        // let request = request.map(|body| body.tee(stderr()).boxed());
+        // let request = request.map(|body| body.tee(io::empty()).boxed());
         //
 
         async {
             let response = next.call(request).await?;
-            Ok(response.map(|body| body.tee(stderr())))
+            Ok(response.map(|body| body.tee(io::empty())))
             //
             // For additional indirection you could alternatively do...
-            // Ok(response.map(|body| body.tee(stderr()).boxed()))
+            // Ok(response.map(|body| body.tee(io::empty()).boxed()))
             //
         }
     });
--- a/examples/echo-server/src/main.rs
+++ b/examples/echo-server/src/main.rs
@@ -22,15 +22,15 @@ async fn echo(request: Request, _: Next) -> via::Result {
 async fn main() -> Result<ExitCode, Error> {
     let mut app = via::app(());
 
-    // Include an error boundary to catch any errors that occur downstream.
-    app.include(error_boundary::inspect(|_, error| {
-        eprintln!("Error: {}", error);
-    }));
-
     app.include(|request: Request, next: Next| {
         let request = request.tee(tokio::io::stderr());
         async { Ok(next.call(request).await?.tee(tokio::io::stdout())) }
     });
+
+    // Include an error boundary to catch any errors that occur downstream.
+    app.include(error_boundary::inspect(|_, error| {
+        eprintln!("Error: {}", error);
+    }));
 
     // Add our echo responder to the endpoint /echo.
     app.at("/echo").respond(via::post(echo));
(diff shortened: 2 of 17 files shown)


examples/echo-server/Cargo.toml
```diff
@@ -5,5 +5,5 @@ edition = "2021"
 
 [dependencies]
 http = "1"
-tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
```
zacharygolba, line 8: This will be reverted before the PR is merged. I was a bit too lazy to write tests this morning and there is nothing wrong with that. I'm trying to get as much done as I can before I either find a job or go bankrupt. In either case, I'll have less time to work on this project. In efforts to get as close to an official, stable 2.0 release in the required time frame, I'm stealing the time that I had budgeted for writing tests under the assumption that once it's confirmed to work, I can probably get some tests implemented quickly with something generated from ai as a starting point.

src/body/tee_body.rs
```diff
@@ -0,0 +1,116 @@
...
+        Self {
+            body: BoxBody::new(body),
+            sink: Box::pin(sink),
+            next: None,
+        }
+    }
+
+    #[inline]
+    pub(crate) fn cap(self) -> BoxBody {
+        // I'm considering bringing back a custom impl of `BoxBody` in the
```
zacharygolba, line 35: This is unfortunate but probably necessary. We're making slightly different assumptions compared to the `http-body-util` crate. We're still maintaining compatibility in that any type from the http-body-util crate that is Unpin can also be used as a `RequestBody` or `ResponseBody` by passing it to our `BoxBody` type rather than the one from the `http-body-util` crate. I would totally use the one from the `http-body-util` crate if there is an `into_inner` method added to the `BoxBody` type. There shouldn't be an issue in doing so because moving the pinned box isn't the same as moving data from outside the pinned pointer. The box itself guarantees a stable memory address and I'd prefer not write a function that can recursively box a request or response body every time it's called.

src/body/tee_body.rs
```diff
@@ -0,0 +1,116 @@
...
+        // that we'll want to project
+        todo!()
+    }
+}
+
+impl Body for TeeBody {
+    type Data = Bytes;
+    type Error = DynError;
+
+    fn poll_frame(
```
zacharygolba, line 46: This is still a rough draft. I'm going to think about this much more. No rush to get this feature out.

src/request/request.rs
```diff
@@ -30,19 +31,44 @@ pub struct Request<T = ()> {
...
+            body: HttpBody::Dyn(map(self.body)),
+            ..self
+        }
+    }
+
+    /// Write the contents of the request body into the provided sink when it
+    /// is read.
+    ///
+    #[inline]
+    pub fn tee(self, sink: impl AsyncWrite + Send + Sync + 'static) -> Self {
```
zacharygolba, line 61: I'm sticking with the plumbing metaphors for now. IMO, they make sense.

src/response/response.rs
```diff
@@ -38,14 +46,34 @@ impl Response {
     ///
     #[inline]
     pub fn map(self, map: impl FnOnce(HttpBody<ResponseBody>) -> BoxBody) -> Self {
-        if cfg!(debug_assertions) && matches!(self.inner.body(), HttpBody::Mapped(_)) {
-            // TODO: Replace this with tracing and a proper logger.
-            eprintln!("calling response.map() more than once can create a reference cycle.");
+        if cfg!(debug_assertions) {
+            prefer_unboxed(self.inner.body());
```
zacharygolba, line 50: We can be smarter about this in the future. This works for now.

src/request/request.rs
```diff
@@ -30,19 +31,44 @@ pub struct Request<T = ()> {
...
+        eprintln!("boxing a request body more than once can create a reference cycle.");
+    }
+}
+
 impl<T> Request<T> {
     /// Consumes the request returning a new request with body mapped to the
     /// return type of the provided closure `map`.
     ///
     #[inline]
     pub fn map(self, map: impl FnOnce(HttpBody<RequestBody>) -> BoxBody) -> Self {
```
zacharygolba, line 38: I think we can provide guidance somewhere in the docs explaining the pros and cons of boxing the Request or Response body. The goal isn't to avoid indirection with the `HttpBody` enum, but rather make it obvious exactly how much indirection there is. I can see many use-cases where I would want to do the following in a map function or a middleware that combines `tee` and `map` to create something like the following:

```rust
HttpBody::Dyn(
    BoxBody::new(
        Box::pin(
            TeeBody::new(
                Box::pin(self.body),
                //       ^^^^^^^^^ RequestBody or ResponseBody
                Box::pin(sink),
            )
        )
    )
)
```

It's somewhat difficult to enforce the composition above with ambiguity of the map function. In part, I'm wondering if the `tee()` method should be removed in favor of examples in the docs of the map function to make the composition obvious.

After reading the example in the echo-server a couple of times, I think I definitely prefer making it obvious whats going on here...

examples/echo-server/src/main.rs
```diff
@@ -27,6 +28,27 @@ async fn main() -> Result<ExitCode, Error> {
         eprintln!("Error: {}", error);
     }));
 
+    // Map the request and response bodies to a boxed body type with 2 levels
+    // of indirection. Also, connect a sink a sync to the request and response
+    // body for the purpose of audit logs (because enterprise software).
+    app.include(|request: Request, next: Next| {
```
zacharygolba, line 34: As you can see here, by removing the tee method and a couple of helper functions to `HttpBody`, we can perform a common task required for many enterprise-grade applications for both the sake of information security and audit logs without having to connect the sink to the actual socket. This approach prevents somebodies JWT and / or session cookie from being stolen by someone with access to the hypothetical audit logs.

We definitely need more documentation on the body module. Including examples that use `http-body-util`. I can see my self doing a couple of transformations with the `BodyExt` trait. Such as calling `flat_map` on `TeeBody` to compress the bytes that are written to the socket without affecting the data that is written to the sink. Lot's of potential here.

src/body/tee_body.rs
zacharygolba: Lot's more work to do here. I'm enjoying taking my time to think about the API that makes the most sense.

examples/echo-server/src/main.rs
zacharygolba: Also this example doesn't work because the echo server already uses a boxed body. I'm going to update the example but share a code snippet below because it's valid for any other use case that exclusively uses `ResponseBody`.

```
app.include(|request: Request, next: Next| {
        let future = next.call(request.map(|existing| {
            BoxBody::new(TeeBody::new(
                BoxBody::new(existing.into_inner()),
                tokio::io::stderr(),
            ))
        }));

        async {
            Ok(future.await?.map(|existing| {
                BoxBody::new(TeeBody::new(
                    BoxBody::new(existing.into_inner()),
                    tokio::io::stdout(),
                ))
            }))
        }
});
```

zacharygolba: On second thought, I like the updated example better with the reintroduction of `TeeBody::cap`.

src/request/request.rs
zacharygolba: If you don't need the tee and only need one 1 level of indirection the following code reads nicely:

```rust
app.include(|request: Request, next: Next| {
    let request = request.map(|body| body.boxed());

    async {
        let response = next.call(request).await?;
        Ok(response.map(|body| body.boxed()))
    }
});
```

src/response/response.rs
zacharygolba: I removed the warning because having 2 pointers can be beneficial in many cases.

src/body/tee_body.rs
zacharygolba: After some consideration, I'm thinking that having the struct wrapper around the `Pin<Box<dyn Body>>` is actually preferred here. Pretty sure it's required for the `TeeBody::cap` fn to work properly.

I'm looking forward to writing some examples that use the BodyExt trait in combination with the concrete types provided by this crate.

I think we're doing a good job keeping the API flexible enough to use with other libraries while promoting best practices and simple solutions to common programing tasks.

src/request/request.rs
zacharygolba: In order to support using TeeBody with a single layer of indirection, I'm modifying the map function to expect an `HttpBody<T>` rather than `BoxBody`. It makes it easier for the caller to decide what to do. I'm also adding a tee method to Request and Response as a short hand way to use a TeeBody with a single layer of indirection. If you prefer to have 2, the following code snippet works:

```rust
// Map the request and response bodies to a boxed body type with 2 levels
// of indirection. Also, connect a sink a sync to the request and response
// body for the purpose of audit logs (because enterprise software).
app.include(|request: Request, next: Next| {
    let request = request.map(|body| {
        let sink = tokio::io::stderr();
        BoxBody::new(TeeBody::new(body.boxed(), sink)).into()
    });

    async {
        Ok(next.call(request).await?.map(|body| {
            let sink = tokio::io::stdout();
            BoxBody::new(TeeBody::new(body.boxed(), sink)).into()
        }))
    }
});
```

src/body/tee_body.rs
```diff
@@ -0,0 +1,111 @@
...
+                // RequestBody and ResponseBody are both Unpin.
+                //
+                // It's possible to have an unsafe pin projection here if you
+                // supply an !Unpin stream to `Pipe::pipe` and then supply a
+                // sink to Response::tee. So long as hyper::body::Incoming is
+                // Unpin,
+                let ready = match Pin::new(&mut this.body).poll_frame(context) {
+                    Poll::Ready(Some(Ok(frame))) => frame,
+                    poll @ (Poll::Pending | Poll::Ready(None) | Poll::Ready(Some(Err(_)))) => {
+                        if this.sink.as_mut().poll_shutdown(context)? == Poll::Pending {
```
zacharygolba, line 64: Currently, the sink is not shutdown for types without an upper bound. This a bug and will be fixed before a review is requested.

src/body/tee_body.rs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            match Pin::new(&mut this.body).poll_frame(context) {
+                Poll::Pending => break Poll::Pending,
+                Poll::Ready(None) => {
+                    done = true;
+                }
+                Poll::Ready(Some(Ok(frame))) => {
+                    if let Some(next) = frame.data_ref() {
+                        backlog.push_back(next.clone());
```
zacharygolba, line 106: This should paradoxically make the implementation zero copy.

zacharygolba:
The easiest examples are always the most dangerous.

Fortunately, danger is my middle name...

examples/hello-world/src/main.rs
```diff
@@ -21,6 +22,13 @@ async fn main() -> Result<ExitCode, Error> {
     // Create a new application.
     let mut app = via::app(());
 
+    app.include(|request: Request, next: Next| async {
+        Ok(next.call(request).await?.map(|body| {
+            let scc = stderr();
+            body.tee(scc).boxed()
```
zacharygolba, line 28: Let us pray that the SCC doesn't have anyone's JWT! 🙏 

The fact that I'm using acronyms means that miracles do happen. So at least there is hope..

zacharygolba: Oh shit I was wrong about being able to use acronyms :scream:

zacharygolba: When the SEC steals a JWT in an audit log, nobody wins.

This isn't a commitment to help people commit financial crimes but rather an attempt to prevent them from getting off the hook for financial crimes.

That's why the multi threaded runtime was implemented first. I care deeply about due process. ✌️

src/body/tee_body.rs
```diff
@@ -81,7 +81,7 @@ impl Body for TeeBody {
                         }
                     }
                     Poll::Ready(Ok(len)) => {
-                        if len < next.len() {
+                        if len < next.remaining() {
```
zacharygolba, line 84: While the implementation of `Buf::remaining()` for `Bytes` delegates directly to  `Bytes::len()`, a goal of this project is auditability. Even ChatGPT thought that this could result in a potential buffer overflow when I asked for a code review.

Using `.remaining()` here removes ambiguity and is what this project is all about.

zacharygolba:
Still have a couple of changes I want to make here before marking the PR as ready for review.

In particular, I would like to make sure that we don't leave cloned handles to data frames in the VecDeque in the event of an error occurring in either `poll_frame` or `poll_write`. This could potentially create a reference cycle with the underlying arc stored in the bytes instance, preventing the memory from being freed in the event of an error.  This should be relatively simple since we have the AsyncWrite shutdown logic in place already.

src/body/tee_body.rs
```diff
@@ -75,11 +74,13 @@ impl Body for TeeBody {
...
                         // Placeholder for tracing...
-                        if polled_frame {
-                            return Poll::Pending;
-                        }
+                        //
+                        // Something along the lines of:
+                        // tracing::info!("TeeBody: io is not yet ready for writes.");
+                        //
+                        backlog.push_front(next);
+                        return Poll::Pending;
```
zacharygolba, line 83: This introduces back pressure in the event that the AsyncWrite is not ready for writes. We're only storing 1 data frame in the backlog while we wait for the socket to become available because the additional capacity in the VecDeque is in place exclusively for partial writes as seen in the match arm for `Poll::Ready(Ok(len))`.

zacharygolba: It's possible but very unlikely that a slower implementation of AsyncWrite could cause the VecDeque in IoState::Writeable to allocate. Rather than trying to prevent this from happening, I think it's best that we add tracing or placeholders for tracing so we know if / when we end up needing to allocate additional capacity for the backlog. This metric will likely be useful in determining the size we end up choosing for `MAX_FRAME_LEN` in ResponseBody. We can slowly move towards `64kb` frames, if we see that TeeBody requires an additional allocation when paired with a TcpStream or an io stream such as stderr, we can adjust accordingly.

src/body/tee_body.rs
```diff
@@ -0,0 +1,163 @@
...
+        let this = self.get_mut();
+
+        let io = &mut this.io;
+        let state = &mut this.state;
+        let backlog = &mut this.backlog;
+        let mut next = None;
+
+        loop {
+            match state {
+                IoState::Writeable => {}
```
zacharygolba, line 61: I'm having a hard time dealing with this empty match arm here. Perhaps basing the state on the shutdown rather than the write-ability of `io` would make my intentions more clear?

I'm going to brainstorm a bit, write some tests, and then mark this as ready for review and merge asap to get another release out along with a couple refactors to the body module in a follow up PR.

zacharygolba:
This is starting to look a lot like the first official crate.

This is a dystopic thing to write and assuming it's always necessary to tee the stream for the sake of auditing sends the wrong message.

It's important to be empathetic to the communities we want to help build software. Such as government, healthcare, finance and others where auditing is needed.

We also don't want to assume that they are our users either.

In other words, I'd like to advertise to them by providing something incredibly useful that's hard to get right at arms reach. I also don't want a casual user stumbling upon it in autocomplete and think that it's something that most applications need.

zacharygolba:
Also it's not finished. Support for an !Unpin Body and AsyncWrite would result in additional improvements to the poll_frame method by improving the layout of the data.

zacharygolba:
Definitely open source Apache 2.0 or MIT. Same repo, cargo workspace, everything

zacharygolba:
If nothing else, it's great practice for a tough problem to solve. Just in time for any code challenges that may or may not get sent my way. I'm available for hire as a software engineer or standup comedian.

pino@hey.com

zacharygolba:
Closing in favor of #166.

PR #171, opened 2025-03-30
feat(via-router): reduce allocations and increase perf

zacharygolba:
Additional router optimizations that didn't make it in to #167.

### DIFF
--- a/via-router/src/router.rs
+++ b/via-router/src/router.rs
@@ -143,6 +143,7 @@ impl<T> Router<T> {
                             next.push(&node.children);
                         }
                     }
+                    Pattern::Root => continue,
                 }
             }
 
@@ -176,7 +177,7 @@ impl<T> Router<T> {
 impl<T> Default for Router<T> {
     fn default() -> Self {
         Self {
-            tree: vec![Node::new(Pattern::Static("".to_owned()))],
+            tree: vec![Node::new(Pattern::Root)],
         }
     }
 }
--- a/via-router/src/router.rs
+++ b/via-router/src/router.rs
@@ -110,7 +110,7 @@ impl<T> Router<T> {
     pub fn visit<'a>(&'a self, path: &str) -> Vec<Binding<'a, T>> {
         let mut segments = Split::new(path).lookahead();
 
-        let mut bindings = Vec::with_capacity(8);
+        let mut bindings = Vec::new();
         let mut branch = Vec::with_capacity(64);
         let mut nodes = Vec::with_capacity(1);
         let mut next = SmallVec::<[&[usize]; 2]>::new();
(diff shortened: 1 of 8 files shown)


via-router/src/router.rs
```diff
@@ -109,75 +108,76 @@ impl<T> Router<T> {
...
-                            nodes.push(Match::new(true, Some(param), route));
-                        }
-                    }
-
-                    results.push(Binding::new(None, nodes));
-                    return results;
-                }
-            };
+        while let Some((range, has_next)) = segments.next() {
+            let mut nodes = Vec::with_capacity(4);
```
zacharygolba, line 125: I plan on coming up with some sort of data structure to get rid of the allocation per path segment that we are doing currently. 

A self-referential struct where binding holds indices to the matched nodes is what I'm thinking. Reducing allocations and increased cache locality is all I would expect from such a refactor.

zacharygolba: I'm thinking this would be a great addition to a follow up PR.

via-router/src/router.rs
```diff
@@ -109,75 +108,80 @@ impl<T> Router<T> {
...
                 let node = lookup!(&self.tree, key);
 
                 match &node.pattern {
                     Pattern::Wildcard(param) => {
                         nodes.push(Match::new(true, Some(param), &node.route))
                     }
                     Pattern::Dynamic(param) => {
-                        next.extend_from_slice(&node.children);
                         nodes.push(Match::new(exact, Some(param), &node.route));
+                        next.push(&node.children);
```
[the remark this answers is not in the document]
zacharygolba: I think? the compiler would generate the same code and I prefer syntax in this case. Please correct me if I'm wrong here.

via-router/src/router.rs
```diff
@@ -223,22 +227,141 @@ where
...
+                    } else {
+                        segment.to_owned()
+                    }
+                })
+                .fold(String::new(), |path, next| path + "/" + &next);
+
+            router.visit(&path);
+        }
 
         // {
```
zacharygolba, line 366: I'm going to fix the tests. Not tonight and perhaps not before the next release. Heads up, empty path segments are a thing now (not a bug). In my opinion, they are completely safe and preferable. In probably every scenario it results in the router returning early rather than assuming it's a typo.

zacharygolba: For now, I think using SmallVec for nodes to optimize for the case of a single match is sufficient...

zacharygolba: I also tried implementing a flat version of the results vec as we had before. I think keeping the tree structure in tact makes sense. Even if it's a little slower, I like the way the results match the structure of the router.

PR #180, opened 2025-04-07
refactor(via-router): prefer a has_nodes flag in Binding

zacharygolba:
Keeps a boolean flag as a cache for whether or not `!self.nodes.is_empty()` in Binding. This change is motivated by the following:

1. Binding should fit in a single cache line (also in the vec) at this size (`64` instead of `56`). This also benefits memory alignment in the results vec returned from `Router::visit`

2. The code instead the for loop in visit has less branching because we no longer have to query whether or not `self.nodes` spilled to the heap

3. We can use a specialized method that no longer has to be negated `self.has_nodes` vs. `!node.is_empty()`

### DIFF
--- a/via-router/src/binding.rs
+++ b/via-router/src/binding.rs
@@ -70,9 +70,7 @@ impl<'a, T> Binding<'a, T> {
     #[inline]
     pub(crate) fn push(&mut self, node: MatchKind<'a, T>) {
         self.nodes.push(node);
-        if !self.has_nodes {
-            self.has_nodes = true;
-        }
+        self.has_nodes = true;
     }
 }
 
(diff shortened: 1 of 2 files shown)


via-router/src/binding.rs
```diff
@@ -47,13 +43,36 @@ impl<T> Binding<'_, T> {
...
+            has_nodes: true,
+            nodes,
+            range,
+        }
     }
 
     #[inline]
     pub(crate) fn push(&mut self, node: MatchKind<'a, T>) {
         self.nodes.push(node);
+        if !self.has_nodes {
```
zacharygolba, line 73: I was debating whether or not I should keep the branch here. I'm leaning towards unconditionally setting has_nodes to decrease the number of instructions

PR #182, opened 2025-04-07
fix(via-router): decrease the branch capacity to 32 

papa-zakari:
Increases the preallocated size of the branch vector to roughly `1kb`. This should result in a single allocation for the branch vec for the vast majority of applications. Even those with many routes in an unbalanced tree.


zacharygolba:
I think I'm going to probably decrease the size back to 64 under the assumption that most trees are balanced in well-designed applications and the example in benches is particularly unbalanced.

zacharygolba:
I decided to favor applications with well-balanced trees even more by reducing the branch capacity to 32. This will look bad in benchmarks. However, the benefit for most applications should be significant.

PR #184, opened 2025-04-07
chore(via): bump version to 2.0.0-rc.32

zacharygolba:
Via v2.0.0-rc.32

# Changelog

- #183


zacharygolba:
Prefix wildcard paths to avoid a dynamic allocation.

PR #190, opened 2025-04-08
refactor(via): use http-body-util as much as possible

zacharygolba:
Cleans up the surface area of the public API in preparation for a code freeze and focus on documentation.


src/body/buffer_body.rs
```diff
@@ -12,30 +13,32 @@ pub const MAX_FRAME_LEN: usize = 8192; // 8KB
...
-            None => Poll::Ready(None),
-            Some(data) => {
-                self.offset += 1;
-                Poll::Ready(Some(Ok(Frame::data(data))))
-            }
-        }
+        let this = unsafe { self.get_unchecked_mut() };
+
+        Poll::Ready(if let Some(next) = this.parts.get(this.offset) {
+            this.offset += 1;
```
zacharygolba, line 33: Don't worry everyone...

If incrementing a struct field with dot notation is considering moving data out from behind a pinned reference, I'll go back to writing ruby and flowtype.

src/middleware/respond_to.rs
```diff
@@ -0,0 +1,110 @@
...
+        predicate: http::Method::TRACE,
+        middleware,
+    }
+}
+
+impl<State, T> Middleware<State> for RespondTo<T>
+where
+    T: Middleware<State>,
+{
+    fn call(&self, request: Request<State>, next: Next<State>) -> FutureResponse {
```
zacharygolba, line 103: A named type is always going to be easier for the compiler to optimize.

Plus, comparing a reference with a reference is going to be faster than dereferencing the method in order to do the comparison by value.

Moving a reference into a closure would create a type that cannot be named. This does what we want and it's definitely not a class. There is no vtable.

src/request/request.rs
```diff
@@ -1,12 +1,30 @@
...
+    /// middleware in the middleware stack for the request, this will be empty.
+    ///
+    cookies: CookieJar,
+
+    /// The request's path and query parameters.
+    ///
+    params: PathParams,
+}
 
 pub struct Request<T = ()> {
```
zacharygolba, line 29: I really like the idea of request and next being able to fit in a single cache line. The allocations required for request are likely hoisted to the top of AppService::call along with the necessary allocation for path params.

src/body/body_reader.rs
```diff
@@ -26,7 +23,7 @@ pub struct BodyData {
 
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub struct BodyReader {
```
zacharygolba, line 25: I prefer the `http_body_util::Collected` struct to `via::body::BodyData` but it cannot be constructed from it's arguments and I prefer `via::body::BodyReader` to `http_body_util::combinators::Collect`.

src/body/buffer_body.rs
zacharygolba: This is of course, Rust but let's break down my thinking in Ruby...

```ruby
class Buf
  attr_accessor :offset
  def initialize
    @offset = 0
  end
end

b = Buf.new
# => #<Buf:0x00000001511d4e60 @offset=0>
b.inspect
# => "#<Buf:0x00000001511d4e60 @offset=0>"
#           ^^^^^^^^^^^^^^^^^^
# The memory address of `b = Buf`
b.offset += 1
# => 1
b.inspect
# => "#<Buf:0x00000001511d4e60 @offset=1>"
#           ^^^^^^^^^^^^^^^^^^
# The memory address of `b` remains the same (and pressumably) 1 now occupies the
# address where 0 used to be.
```

via-router/src/router.rs
```diff
@@ -43,17 +43,44 @@ impl<T> Node<T> {
...
         }
     }
 
     #[inline]
     pub fn route(&self) -> slice::Iter<MatchCond<T>> {
         self.route.iter()
     }
+
+    #[inline]
+    pub fn matches<'node, 'predicate, U>(
```
zacharygolba, line 63: This MatchCond enum is way to fun to play around with. In practice, it's probably better to unwrap the contained ref to node if you use params like we do in `AppService::call`.

src/body/body_stream.rs
```diff
@@ -1,49 +1,47 @@
...
-    pub(crate) fn new(body: HttpBody<RequestBody>) -> Self {
+    pub(crate) fn new(body: BoxBody) -> Self {
         Self { body }
     }
 }
 
+impl BodyStream {
+    #[inline]
+    fn project(self: Pin<&mut Self>) -> Pin<&mut BoxBody> {
+        Pin::new(&mut self.get_mut().body)
```
zacharygolba, line 27: Desugared:

```rust
// This is probably fine...
Pin {
    __pointer: &mut self.__pointer.body,
}
```

src/response/response.rs
```diff
@@ -31,41 +35,38 @@ impl Response {
     #[inline]
     pub fn map<F>(self, map: F) -> Self
     where
-        F: FnOnce(HttpBody<ResponseBody>) -> HttpBody<ResponseBody>,
+        F: FnOnce(Either<BufferBody, BoxBody>) -> Either<BufferBody, BoxBody>,
     {
         Self {
-            inner: self.inner.map(map),
+            inner: self.inner.map(|body| ResponseBody {
```
zacharygolba, line 41: It's a good thing the response body isn't polled until it's unwrapped from `Response` on line 87 in `crate::app::service`.

src/request/request.rs
```diff
@@ -226,7 +241,16 @@ impl<T> Debug for Request<T> {
             .field("headers", self.headers())
             .field("params", &self.head.params)
             .field("cookies", &self.head.cookies)
-            .field("body", self.body())
+            .field("body", &self.body)
             .finish()
     }
 }
+
+impl<T> Pipe for Request<T> {
```
zacharygolba, line 249: This literally exists only for the sake of having a terse echo-server example.

The good news is that `Pipe` no longer requires `Sealed` so users of via will be able to impl `Pipe` for external types.

src/body/buffer_body.rs
zacharygolba: To my understanding, pinning a reference doesn't mean `mem::forget`. The struct itself is pinned, the value of each field  will occupy the same space in memory and contain the same types. 

An internal pointer "moves" to the next offset but does not move out of self.

If this violates the implied contract of an `!Unpin` type, how can anyone implement an `!Unpin` body when you have to have some way of knowing whether or not to return `Poll::Ready(None)`?

I'm going to re-implement the version that used a single bytes instance that uses the slice method as I think that's a nicer algorithm and the memory is guaranteed to be contiguous that way. That will still require decrementing a remaining field that will require unsafe code to implement as `!Unpin`.

zacharygolba: It seems like the community tends to prefer `Unpin` types anyways. 

I'm going to make the change I mentioned using `Bytes::slice` method and remove the `_pin` field.

Admittedly, it's been a while since I read something from the rustonomicon.

Maybe there is a section about advanced pin techniques that would make the simple design decision I'm trying to make a bit easier by showing significantly more complex examples?

I'm available for work and miss having coworkers.

pino@hey.com

zacharygolba: FWIW I love both ruby and flowtype. ❤️

src/server/accept.rs
```diff
@@ -132,19 +130,20 @@ where
...
                         // Create a new HTTP/1.1 connection.
                         #[cfg(all(feature = "http1", not(feature = "http2")))]
-                        let mut connection = conn::http1::Builder::new()
-                            .timer(TokioTimer::new())
-                            .serve_connection(IoStream::new(accepted), service)
-                            .with_upgrades();
-
-                        // Pin the connection on the stack so it can be polled.
-                        let mut connection = Pin::new(&mut connection);
+                        let mut connection = Box::pin(
```
zacharygolba, line 141: I've seen a noticeable improvement with regards to predictable memory usage after making this change.

zacharygolba: Ended up changing my mind back. If we're going to allocate once in the connection task before hitting an await point, we should use it for the socket. That way, we know that it's memory address will not change under any circumstances.

src/server/accept.rs
```diff
@@ -0,0 +1,194 @@
...
+        // Acquire a permit from the semaphore.
+        let permit = match semaphore.clone().acquire_owned().await {
+            Ok(acquired) => acquired,
+            Err(_) => break 1.into(),
+        };
+
+        // Accept the next stream from the tcp listener.
+        let (stream, _addr) = tokio::select! {
+            biased;
+            result = time::timeout(Duration::from_secs(15), listener.accept()) => {
```
zacharygolba: I really like this idea and can see it being implemented in the future if the community agrees with me.

This seems a bit novel and I would like to get some feeback from everyone before exploring this approach further.

PR #191, opened 2025-04-11
feat(via): cleanup the accept loop

zacharygolba:
Garbage collect when the server is idle, join a connection before continuing to the next iteration when the server is at capacity.


zacharygolba:
The description of this PR isn't great. The design I have in my head works similar to a transmission where the server could essentially downshift if it's under load.

This is very much an experiment and is not guaranteed to be the next iteration.

src/server/accept.rs
```diff
@@ -64,7 +52,11 @@ where
     // of connections that the server can handle concurrently. If the maximum
     // number of connections is reached, we'll wait until a permit is available
     // before accepting a new connection.
-    let semaphore = Arc::new(Semaphore::new(max_connections));
+    let semaphore = Arc::new(Semaphore::new(context.max_connections()));
+
+    // Create a JoinSet to track inflight connections. We'll use this to wait for
+    // all connections to close before the server exits.
+    let mut connections = VecDeque::<TaskHandle>::with_capacity(4096);
```
zacharygolba, line 59: I'm a fan of the big allocation here with `VecDeque`. I think we're smart enough to guess the allocation size required to support the server even if it's at capacity.

This is one of the only reasons why I would prefer using a VecDeque rather than `JoinSet`. The implementation of `JoinSet` is great but I'd prefer that no dynamic allocations occur in the accept loop.

I'm considering submitting a PR to tokio that adds a `with_capacity` method to `JoinSet` for this reason.

src/server/accept.rs
```diff
@@ -89,83 +74,111 @@ where
...
-                Ok(accepted) => accepted,
-                Err(error) => {
-                    eprintln!("error(listener): {}", error);
-                    drop(permit);
-                    continue;
+        let (stream, _addr) = {
+            let accept = poll_fn({
+                let listener = &listener;
+                let connections = &mut connections;
+                let mut interval = time::sleep(GC_INTERVAL);
```
zacharygolba, line 84: This is where it starts to feel over-engineered.

One of the things I love about using `JoinSet` is that `join_next` is cancel safe.

We are already using `tokio::select` to poll for the next connection or shutdown signal concurrently. Switching back to the original design goes as follows:

1. Wait for either a new incoming connection, the next connection to be joined, or a shutdown signal.

    a. If a connection is joined before a new one arrives, continue to the next iteration of the select loop.

    b. If a shutdown signal is received, stop accepting new connections and attempt to join each inflight connection within `shutdown_timeout`.

2. A new connection is accepted, the connection we attempted to join will remain at the front of the `JoinSet`. Spawn a new connection task to handle the connection.

    a. This behavior is desirable so long as the connection is not used to create a websocket. Otherwise, a deque-like data structure is preferred to de-prioritize these types of connections so they can be moved to back of the queue.

3. Check and see if the connection we attempted to join in the select loop is finished using `JoinSet::try_join_next`.

4. Continue to the next iteration of the accept loop.

---

In the absence of a public websocket API, the only thing I would consider changing about this approach is adding a step between 2 and 3 to maybe yield on the main thread of the accept loop for a single turn to give connections a little more CPU-time.

zacharygolba: The nice thing about the approach I mention above is that there is less contention by relying less on timers expiring.

src/server/accept.rs
```diff
@@ -89,83 +74,111 @@ where
...
-
-                        // Serve the connection.
-                        tokio::select! {
-                            result = &mut connection => result.map_err(|e| e.into()),
-                            _ = shutdown_rx.changed() => {
-                                connection.as_mut().graceful_shutdown();
-                                connection.await.map_err(|e| e.into())
-                            }
-                        }
+            // TODO: Attempt to get the task size in the tls example <= 1024.
```
zacharygolba, line 131: This is probably something we want to be cognizant of.

It's probably safe to consider anything under `2048` a reasonable task size for an HTTPS server. If my memory serves me correctly, that is the norm.

It's also somewhat likely that a smaller task size in general is more desirable than avoiding an allocation in the accept loop. Especially if our intention is to yield to give the task we attempted to join in the select loop time to finish.

I'm unlikely to make a decision here until I perform some serious benchmarks and profiling on Linux. I've been developing exclusively on macOS as I do a lot of React Native work professionally.

zacharygolba: FWIW, I have a hunch that app servers should always prefer the latter and the numbers I've been coming up with are default values that I saw countless times in apache / nginx config files.

src/server/accept.rs
```diff
@@ -89,83 +74,111 @@ where
...
+            task::spawn(async move {
+                let io = match acceptor.accept(stream).await {
+                    Ok(accepted) => TokioIo::new(accepted),
```
zacharygolba, line 134: We probably want to wrap the socket as late as possible here.

I understand the important of needing writes to the socket to be as fast as possible.

Therefore, I'm thinking we stick with `TokioIo` instead of `IoStream` and consider boxing the connection rather than the task or the socket. It should result in a predictable task size that is small enough to offer greater scalability but not so compact that we potentially overflow the stack.

zacharygolba: The correct approach here may vary based on the runtime that is used.

src/server/accept.rs
```diff
@@ -89,83 +74,111 @@ where
...
                 result
-            }
+            })
         });
 
-        for _ in 0..CONNECTION_JOIN_LIMIT {
-            match connections.try_join_next() {
-                Some(result) => joined!(&result),
-                None => break,
+        if semaphore.available_permits() == 0 {
```
zacharygolba, line 172: I really like the idea of requiring that a connection be joined if the server is at capacity.

However, querying the low-level semaphore is probably something we want to avoid.

If we decide to implement this in the future, it's probably best to track this state ourselves.

src/server/accept.rs
```diff
@@ -176,13 +189,83 @@ where
     }
 }
 
-async fn drain_connections(connections: &mut JoinSet<Result<(), DynError>>) {
+fn joined(result: &Result<TaskResult, JoinError>) {
```
zacharygolba, line 192: This isn't pretty but it works just fine.

src/server/accept.rs
```diff
@@ -176,13 +189,83 @@ where
...
+                    // trace!();
+                } else {
+                    log!("error(http): {}", e);
+                }
+            } else {
+                log!("error(other): {}", error);
+            }
+        }
+        // The connection task panicked.
+        Err(error) if error.is_panic() => panic!("{}", error),
```
zacharygolba, line 215: If a task panics due to a poisoned mutex, we definitely want to panic on the main thread.

This seems less resilient but this is Rust, not Java. Restarting the main process from a supervisor is a best practice and one we should assume people are doing. This is what we would want regardless if we implemented the single threaded version of the server, first.

src/server/acceptor/rustls.rs
```diff
@@ -22,10 +26,10 @@ impl RustlsAcceptor {
 }
 
 impl Acceptor for RustlsAcceptor {
-    type Future = Accept<TcpStream>;
-    type Stream = TlsStream<TcpStream>;
+    type Stream = Box<TlsStream<TcpStream>>;
```
zacharygolba, line 29: I'll probably find a better way to keep the size of tls streams below 2048. For now, this is fine. It's not like people are filling up the backlog with issues complaining about poor performance when tls is in use.

zacharygolba:
Aside from using a hash map for path params, the big allocation and the accept loop changes in this pr are the only things preventing me from finding some pen testers on fiverr to try to poke some holes here.

There is a performance regression that I'm seeing locally but I'm less concerned about that.

src/server/conn.rs
```diff
@@ -49,9 +49,17 @@ impl JoinQueue {
             joined = self.join_next() => joined,
             _ = time::sleep(timeout) => {
                 if self.queue.len() > 1 {
-                    self.queue.swap(0, 1);
```
zacharygolba: Rather than moving the join handle to the next position in the queue, we move it to the back. This should always result in less contention. Especially if the connection that could not be joined within the deadline uses a websocket.

src/server/conn.rs
```diff
@@ -49,9 +49,17 @@ impl JoinQueue {
...
+                    // If the connection task can't be joined in `timeout`,
+                    // move it to the back of the queue. It might be using
+                    // a websocket.
+                    match self.queue.pop_front() {
+                        Some(handle) => self.queue.push_back(handle),
+                        None => panic!("join_next was not canceled safely"),
```
zacharygolba, line 57: The implementation of `join_next` is potentially not cancel safe. A little bit of work is required to ensure that we are joining the same join handle when the task is woken up by the waker.

Perhaps, the `idle` field can be changed to an `Option<Id>` to ensure that the same task will be polled when the waker wakes up the task to continue making progress.

From what I can tell, there is little risk associated with this in the current implementation. It seems like the worst case scenario would be polling a different join handle at the front of the queue if `wake_by_ref` is called just before the timeout expires.

PR #193, opened 2025-09-01
feat(via-router): return an iter from Router::visit

zacharygolba:
Prefer returning bindings to each path segment as an iterator rather than a Vec to reduce allocations. This should improve both the security and performance of a critical path.

The API needs cleaned up, comments have to be added, and tests have to be fixed.


via-router/src/router.rs
```diff
@@ -115,196 +87,149 @@ impl<T> Router<T> {
...
     ///
     /// If a node referenced by another node does not exist in the route tree.
     /// This router is insert-only, therefore this is a very unlikely scenario.
     ///
-    pub fn visit<'a>(&'a self, path: &str) -> Result<Vec<Binding<'a, T>>, Error> {
-        let mut segments = Split::new(path).lookahead();
-        let mut results = Vec::with_capacity(VISIT_RESULTS_CAPACITY);
-        let mut branch = Vec::with_capacity(VISIT_BRANCH_CAPACITY);
-        let mut next = SmallVec::<[&[usize]; 2]>::new();
+    pub fn visit<'a, 'b>(&'a self, path: &'b str) -> impl Iterator<Item = Binding<'a, T>> + 'b
```
zacharygolba, line 105: Refactoring this fn is the focus of the PR.

src/app/service.rs
```diff
@@ -38,31 +37,33 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
...
-                params.extend(match kind {
-                    MatchKind::Edge(MatchCond::Partial(node)) => {
-                        next.extend(node.route().filter_map(MatchCond::as_partial).cloned());
-                        node.param(|| binding.range())
-                    }
-                    MatchKind::Edge(MatchCond::Exact(node)) => {
-                        next.extend(node.route().map(MatchCond::as_either).cloned());
-                        node.param(|| binding.range())
+        for binding in self.app.router.visit(path) {
+            for node in binding.results() {
```
zacharygolba, line 46: While returning references to nodes from via-router could potentially expose the entire route tree to a hacker, the references are short lived enough to not make a difference.

If anyone disagrees, please comment and provide an explanation.

As far as I can tell, as long as the references are not live between await points in the code, you would have to be the luckiest person on the planet to get a reference back to the route tree while the server is still running.

src/app/service.rs
```diff
@@ -40,47 +39,28 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
...
-                    _ => binding.is_final(),
-                };
-
-                if match_as_final {
-                    // Include the entire route stack of the matched node.
-                    next.extend(node.as_final().cloned());
-                } else {
-                    // Only include the middleware of the matched node.
-                    next.extend(node.as_partial().cloned());
+        for binding in self.app.router.visit(request.uri().path()) {
```
zacharygolba, line 44: The latest commit moves the responsibility of filtering route handlers from the API consumer of via-router into via-router. The result is a cleaner public API and likely better performance. Currently, wildcard path ranges are not accounted for. I'm going to make a change that fixes that after a quick lunch break.

src/middleware/next.rs
```diff
@@ -1,17 +1,17 @@
-use std::collections::VecDeque;
+use smallvec::SmallVec;
 use std::sync::Arc;
 
 use super::middleware::{BoxFuture, Middleware};
 use crate::error::Error;
 use crate::request::Request;
 
 pub struct Next<T = ()> {
-    deque: VecDeque<Arc<dyn Middleware<T>>>,
```
zacharygolba, line 9: For the sake of correctness, I think I'm going to drop this commit and preallocate for a VecDeque even though it's less performant.

In theory, there isn't anything wrong with this optimization. However, it does feel like implementing a middleware stack cache similar to Actix is a better approach. I prefer to an extra allocation until we optimize the right way.

via-router/src/router.rs
```diff
@@ -45,21 +45,21 @@ fn match_next_segment<'a, T>(
     segment: &str,
     range: [usize; 2],
 ) -> Option<Binding<'a, T>> {
-    let mut results: Option<SmallVec<_>> = None;
```
zacharygolba: While this seems like a reasonable optimization, I prefer a length check to determine whether or not we should return `Some` or `None`.

In a vacuum, there is no reason why that null pointer optimization would take away from the resilience of the router. However, if you factor in entropy, it's probably best if we do a len check.

Since `SmallVec` checks whether or not the inline storage spilled to the heap, it's slightly more complicated than checking a len ptr and as a result–it makes it more difficult to "trick" the router into thinking that the binding is empty.

PR #203, opened 2025-09-07
refactor(via): http error ergonomics

zacharygolba:
Ergonomic improvements to the `Error` struct.

- `Error::new` now takes a `StatusCode` argument
- Status code based constructors are now generic (opaque boxing)
- Introduce a `format!`-like error macro with a status code argument

These changes will make a pretty big difference in user land. Being able to `.map_err(Error::bad_request)` without using a closure to box the original error is great. If you already have a boxed error source, the `Error` constructor can be used instead.


zacharygolba:
Docs can be added in a follow up PR.

PR #224, opened 2025-09-11
feat(via): work in progress impl of ws support

zacharygolba:
Work in progress of basic websocket support. We don't want to recreate phoenix. We also don't want to prevent anyone from using Via to do so.

Still a work in progress. It's a bit of a mess but it's working for the most part.


zacharygolba:
This can probably be improved and definitely expanded on. I don't think the implementation will change much from a public API perspective though. Maybe we'll expose a trait for reading client data and return an opaque type instead of RequestPayload when reading a request body and maybe a generic impl for T where T: Buf. The only thing beneficial it would provide on top of the current api would probably be a parse_json method. Not a breaking change though and I want to get in the habit of thinking about that.

If anyone has any comments please feel free to retroactively review this.

PR #228, opened 2025-09-17
feat(via): 24 hr timeout on accept if connections.is_empty()

### LINKED ISSUE #226: Make the timeout on accept conditional
Right now, we're aggressively joining connections by timing out on accept in order to free resources under the assumption that they are necessary for the next turn of the accept loop. I like this idea a lot. It's probably exactly what you want when you're at or near the maximum number of concurrent users that your API can handle. You always have room for another.

It would be relatively easy to make the timeout conditional by including a bit of book keeping in an `AtomicUsize`. It's probably a good idea to avoid checking the number of available permits in the semaphore and the 2 additional atomic operations will probably help synchronize things a bit. It's probably a good idea to await an accept if the server is idle in the event that the hardware that it's running has any energy saving features.

zacharygolba:
Closes #226

### DIFF
--- a/src/server/accept.rs
+++ b/src/server/accept.rs
@@ -18,6 +18,17 @@ use super::io::IoWithPermit;
 use crate::app::{App, AppService};
 use crate::error::ServerError;
 
+macro_rules! accept_with_timeout {
+    ($future:expr, $immediate:expr) => {
+        time::timeout(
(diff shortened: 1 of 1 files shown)


zacharygolba:
We can make this more sophisticated in the future. For now, I think this does a pretty good job at incrementally cleaning resources. Also, we can adjust this as needed as it's not a part of the public API.

PR #239, opened 2025-09-20
feat(via-router): avoid allocs in happy path

zacharygolba:
Reintroduces a happy path to via-router that avoids allocations if there is only a single match for a given path segment. Also inlines some nested pattern matching that should eliminate a couple of branches.

### DIFF
--- a/src/app/service.rs
+++ b/src/app/service.rs
@@ -95,7 +95,7 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
         // segment.
         for (route, param) in self.app.router.traverse(parts.uri.path()) {
             // Add the matched route's middleware to the call stack.
-            next.extend(route.map(Arc::clone));
+            next.extend(route.cloned());
 
             if let Some((name, range)) = param {
                 // Include the resolved dynamic parameter in params.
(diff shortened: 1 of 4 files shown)


src/app/service.rs
```diff
@@ -95,7 +95,7 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
         // segment.
         for (route, param) in self.app.router.traverse(parts.uri.path()) {
             // Add the matched route's middleware to the call stack.
-            next.extend(route.map(Arc::clone));
+            next.extend(route.cloned());
```
zacharygolba, line 98: Using `cloned()` eliminates the need to carry around a function pointer to `Arc::clone`.

via-router/src/router.rs
```diff
@@ -45,39 +45,25 @@ struct Node<T> {
...
-    let (value, range) = segment;
     let next = branch.path.next();
+    let mut binding = Binding {
+        is_final: next.is_none(),
+        results: SmallVec::new(),
+        range: Some(range),
+    };
 
-    for node in branch.children {
+    for node in branch.children.iter().rev() {
```
zacharygolba, line 60: It's necessary to iterate in reverse since we're call `.pop()` on `SmallVec` instead of next on an iterator.

PR #243, opened 2025-09-20
feat(via): simple batch joins in accept

zacharygolba:
Avoids doing work on the main thread of accept by joining connections in the background every 1024 connections.

### DIFF
--- a/src/server/accept.rs
+++ b/src/server/accept.rs
@@ -1,12 +1,13 @@
 use hyper::server::conn;
 use hyper_util::rt::TokioTimer;
 use std::error::Error;
+use std::mem;
 use std::process::ExitCode;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


src/server/accept.rs
```diff
@@ -172,14 +159,20 @@ fn handle_error(error: ServerError) {
...
-        if let Err(error) = &result {
-            handle_error(ServerError::Join(error));
+        match result {
+            Ok(Err(error)) => handle_error(error),
+            Err(error) => handle_error(ServerError::Join(error)),
+            _ => {}
+        }
+
+        if !immediate {
+            coop::consume_budget().await;
```
zacharygolba, line 175: When we're joining in the background, yield to the runtime if we exhausted our budget.

PR #251, opened 2025-09-22
refactor(via): builtin middleware can be named

papa-zakari:
Refactors the middleware located within the builtin mod so that each type can be named. This allows middleware that wraps another middleware to delegate to the wrapped middleware without the overhead of a box + vtable.


examples/cookies/src/main.rs
```diff
@@ -106,7 +106,7 @@ async fn main() -> Result<ExitCode, BoxError> {
     // The CookieParser middleware can be added at any depth of the route tree.
     // In this example, we add it to the root of the app. This means that every
     // request will pass through the CookieParser middleware.
-    app.include(cookies::parse_encoded());
+    app.include(cookies::percent_decode());
```
zacharygolba, line 109: This change is for consistency with the path and query params API.

src/builtin/allow.rs
```diff
@@ -0,0 +1,130 @@
+use http::header::ALLOW;
+use http::{HeaderValue, Method};
+
+use crate::middleware::{BoxFuture, Middleware};
+use crate::{Next, Request, Response};
+
+pub struct Allow<State> {
```
zacharygolba, line 7: Allow is a more descriptive name in this particular use case.

PR #283, opened 2025-09-26
refactor(via): flat modules and some name changes

zacharygolba:
Moves builtin middleware out of the builtin module and moves it to the crate root. Via is about simplicity. I don't anticipate or want to ship with a ton of middleware.

It's a lot harder to reason about the basics when we start forming opinions about form validation. I want to leave as much space in the community as possible to fill in the blanks where Via leaves off.

The only other middleware I would anticipate adding is a generic auth middleware that calls a function with a ref to request to get a future that either resolves with a type that populates a generic, optional `user` field on request.

We can of course change direction based on the wants and needs of the community but I think it's important that we keep the scope of the core framework limited because that is in part what makes it special.


zacharygolba:
It's like ordering at a restaurant. We'll let The Cheesecake Factory be The Cheesecake Factory and focus on making every dish on our menu taste fucking incredible.

Screenshot 2025-09-26 alle 5 22 26 PM

PR #292, opened 2025-09-28
feat(via): smallvec optimization for single frame payloads

papa-zakari:
If a request payload consists of a single frame, don't allocate.


zacharygolba:
This is a pretty cool optimization but I'll probably leave things the way they are. If you're reading a body, you have to allocate. It's a confirmation that the integrity of the hardware that you're running on is intact. It's just an extra allocation. Completely fair and reasonable considering we don't allocate until the channel used for the body starts producing frames.

PR #293, opened 2025-10-17
refactor(via): router api name changes

papa-zakari:
Refactors the public API of the router DSL to prefer making the purpose of the function obvious. The result is a slightly less cutesy but easier to grasp what's going on behind the scenes at a glance.

The only functional change is that `Allow::or_next` is now `Allow::or_not_allowed`. This makes the default 405 response for requests with an unsupported method opt-in rather than opt-out. This change is congruent with the "no surprises" mentality of the framework.

### DIFF
--- a/via-router/Cargo.toml
+++ b/via-router/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "via-router"
-version = "3.0.0-beta.28"
+version = "3.0.0-beta.29"
 authors = ["Zachary Golba <zachary.golba@postlight.com>"]
 edition = "2024"
 license = "MIT OR Apache-2.0"
--- a/via-router/benches/bench.rs
+++ b/via-router/benches/bench.rs
@@ -116,7 +116,7 @@ fn find_matches_1(b: &mut Bencher) {
     let mut next = VecDeque::with_capacity(8);
 
     for path in ROUTES {
-        let _ = router.at(path).include(Arc::new(()));
+        let _ = router.at(path).middleware(Arc::new(()));
     }
 
     b.iter(|| {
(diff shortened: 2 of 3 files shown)


examples/blog-api/src/main.rs
```diff
@@ -21,57 +23,51 @@ async fn main() -> Result<ExitCode, BoxError> {
...
-
-        // Define the /api/users resource.
-        api.at("/users").scope(|users| {
-            users.respond(via::get(users::index).and(via::post(users::create)));
+    let mut api = app.route("/api");
+
+    // Capture errors that occur in the api namespace, log them, and then
+    // convert them into json responses. Upstream middleware remains
+    // unaffected and continues execution.
+    api.middleware(rescue(with_error_sanitizer));
```
zacharygolba: It's a lot easier to guess the purpose of `.middleware` rather than `.include`. Especially in the context of `.respond` being called on a subsequent line.

examples/blog-api/src/main.rs
```diff
@@ -21,57 +23,51 @@ async fn main() -> Result<ExitCode, BoxError> {
...
-                via::get(posts::show)
-                    .and(via::patch(posts::update))
-                    .and(via::delete(posts::destroy)),
-            );
-        });
-
-        // Define the /api/users resource.
-        api.at("/users").scope(|users| {
-            users.respond(via::get(users::index).and(via::post(users::create)));
+    let mut api = app.route("/api");
```
zacharygolba: `{app,route}.at` reads very well but it starts to feel a little too much like a DSL. `.route` is obvious and allows for the variable name `at` to be used more frequently in the code for path parameter ranges and other types that point to a location without feeling like the name is overloaded.

zacharygolba: Also, `api` is defined as a variable rather than a scope because scope takes ownership of and consumes the route.

This works great so long as there is not more than one route with nested endpoints that are defined at the top level of the parent route.

PR #302, opened 2025-10-20
refactor(via): AppService semantic improvements

papa-zakari:
Semantic improvements to the implementation of `AppService::call`. This shouldn't make a significant difference in performance or security but it makes the code read better from a semantic perspective.

### DIFF
--- a/src/app/service.rs
+++ b/src/app/service.rs
@@ -84,8 +84,8 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
         // Get a reference to the component parts of the request as well as a
         // mutable reference to the path parameters.
         let RequestHead {
-            ref parts,
             ref mut params,
+            ref parts,
             ..
         } = *request.head_mut();
 


src/app/service.rs
```diff
@@ -47,6 +49,9 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
     type Response = http::Response<ResponseBody>;
 
     fn call(&self, request: http::Request<Incoming>) -> Self::Future {
+        // The middleware stack.
+        let mut deque = VecDeque::new();
```
zacharygolba: It's probably better to move the `VecDeque` into `Next` as soon as the middleware stack is built.

This way, the only thing that you can do with the middleware stack in user-land is call the next middleware.

src/app/service.rs
```diff
@@ -78,49 +83,39 @@ impl<T: Send + Sync> Service<http::Request<Incoming>> for AppService<T> {
...
+                // Include the route's dynamic parameter in params.
                 params.push(Arc::clone(name), range);
             }
         }
 
-        // Call the middleware stack to get a response.
-        ServeRequest(next.call(request))
+        ServeRequest {
+            // Call the middleware stack to get a response.
+            future: Next::new(deque).call(request),
```
zacharygolba: The scope of what can be done with the middleware stack is reduced to `Next::call`. Exactly what we want.

src/app/service.rs
zacharygolba: To reiterate this, at least for the middleware stack–allocation entropy is preferred to preallocating. It's a small allocation that likely comes with zero runtime cost.

PR #305, opened 2025-10-22
refactor(via): usability improvements for Timeout

papa-zakari:
Improves the usability of the `Timeout` middleware by providing the ability to fully customize the response generated when a timeout expires.

Also provides additional convenience methods for creating timeouts and changing the default status code used when a timeout expires.

### DIFF
--- a/src/timeout.rs
+++ b/src/timeout.rs
@@ -8,8 +8,6 @@ use crate::{Error, Next, Request};
 
 /// Enforce that downstream middleware respond within a specified duration.
 ///
-/// If the timeout expires before a response is generated, an error is returned.
-///
 /// # Example
 ///
 /// ```
@@ -46,7 +44,7 @@ impl Clone for OrElse {
 }
 
 impl Timeout {
-    /// Create a `Timeout` middleware from the duration provided.
+    /// Returns a `Timeout` middleware with the provided duration.
     ///
     /// # Example
     ///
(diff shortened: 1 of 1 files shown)


examples/blog-api/src/main.rs
```diff
@@ -44,7 +43,7 @@ async fn main() -> Result<ExitCode, Error> {
     // Add a timeout middleware to the /api routes. This will prevent the
     // server from waiting indefinitely if we lose connection to the
     // database. For this example, we're using a 10 second timeout.
-    api.middleware(timeout(Duration::from_secs(10)));
+    api.middleware(Timeout::from_secs(10).or_service_unavailable());
```
zacharygolba: Constructor syntax for middleware makes the most sense. The only reasonable exception is the `Allow` middleware.

Preferring constructor syntax gives middleware authors the ability to write convenience methods such as `Timeout::from_secs`.

src/timeout.rs
```diff
@@ -1,29 +1,181 @@
...
+///
 pub struct Timeout {
     duration: Duration,
+    or_else: OrElse,
 }
 
-/// Create a new `Timeout` middleware with the specified duration.
-pub fn timeout(duration: Duration) -> Timeout {
-    Timeout { duration }
+enum OrElse {
```
zacharygolba: Implemented as an enum to enforce mutual exclusivity in the type system.

Otherwise, we'd have to override the status code of the response returned from the or_else callback which is redundant if a user is expected to return a result that contains a Response.

PR #306, opened 2025-10-22
refactor(via): ergonomic improvements to Rescue middleware

papa-zakari:
Paradoxically improves the ergonomics of the `Rescue` middleware by not returning `&mut Self` from the methods exposed by the `Sanitizer` API.

The `Sanitizer` API uses mutability rather than the builder pattern. Returning `&mut Self` from the setters on `impl Sanitizer` encourages an anti-pattern. This change makes the assertion made in #300 an explicit part of the way `Sanitizer` works.

### DIFF
--- a/examples/blog-api/src/api/util.rs
+++ b/examples/blog-api/src/api/util.rs
@@ -2,14 +2,13 @@ use diesel::result::{DatabaseErrorKind, Error as DieselError};
 use http::StatusCode;
 use via::error::Sanitizer;
 
-/// Sanitizes details about database errors that
-pub fn with_error_sanitizer(sanitizer: &mut Sanitizer) {
+pub fn error_sanitizer(sanitizer: &mut Sanitizer) {
     // Print the original message to stderr. In production you probably want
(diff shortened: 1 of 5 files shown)


examples/blog-api/src/api/util.rs
```diff
@@ -2,14 +2,13 @@ use diesel::result::{DatabaseErrorKind, Error as DieselError};
...
-/// Sanitizes details about database errors that
-pub fn with_error_sanitizer(sanitizer: &mut Sanitizer) {
+pub fn error_sanitizer(sanitizer: &mut Sanitizer) {
     // Print the original message to stderr. In production you probably want
     // to use env_logger, tracing, or something similar.
     eprintln!("error: {}", sanitizer);
 
     // Configure the sanitizer to generate a JSON response.
-    sanitizer.respond_with_json();
+    sanitizer.use_json();
```
zacharygolba: `Rescue::with` reads much better than `Rescue::new`. In the context of `respond_with_json`, it would become redundant. `set_*` and `use_*` make more sense in this particular context.

PR #307, opened 2025-10-23
docs(via): documentation for the Next struct

papa-zakari:
Documents the `Next` struct in the spirit of this framework. Simplicity is a requirement, not a goal.

### DIFF
--- a/src/next.rs
+++ b/src/next.rs
@@ -4,22 +4,36 @@ use std::sync::Arc;
 use crate::middleware::{BoxFuture, Middleware};
 use crate::request::Request;
 
-pub struct Next<T = ()> {
-    deque: VecDeque<Arc<dyn Middleware<T>>>,
+/// The next middleware in the logical call stack of a request.
+///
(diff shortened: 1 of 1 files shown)


src/next.rs
```diff
@@ -4,22 +4,36 @@ use std::sync::Arc;
 use crate::middleware::{BoxFuture, Middleware};
 use crate::request::Request;
 
-pub struct Next<T = ()> {
-    deque: VecDeque<Arc<dyn Middleware<T>>>,
+/// The next middleware in the logical call stack of a request.
```
zacharygolba: `Next` is not a call stack but it's behavior is identical to one. I think this metaphor is helpful. It allows the reader to conceptualize how a type can be used for control flow at a glance.

Context is also important. This is a web framework and not a portable JavaScript runtime. It would probably be more surprising if Next was implemented like a call stack. I'm open to alternatives but I really like the way that this reads. Looking forward to bike shedding with co-workers about this stuff. Hopefully in the not so distant future.

PR #310, opened 2025-10-24
refactor(via): Pipe ~> Finalize

papa-zakari:
"Pipe" works conceptually if you think in terms of flow composition (values flowing from handler -> middleware -> response), but once you start thinking about branching, finalization, or response construction, the mental model shifts from “data streaming” to "builder completion.".

The `Pipe` trait was first introduced in order to write the [`echo` example](https://github.com/zacharygolba/via/blob/bc2328340df0a2d375c6a7c870889e84fe98acbb/examples/echo-server/src/main.rs). At the time, the only implementers of `Pipe` were `Request` and a generic impl for `T` where `T: Stream<Item = Result<Frame<Bytes>, BoxError>> + Send + Sync + 'static`. The purpose of the trait has since evolved from a means of connecting a producer to a consumer to an alternative to providing an `IntoResponse` trait. It's API is tangential to that of the `ResponseBuilder` and it's name and location now reflect this relationship.

Unlike many popular Rust frameworks, Via chooses not to provide an implicit conversion for types that implement a hypothetical `IntoResponse` trait. This is intentional and is done so to encourage an explicit, concrete return type for all implementers of `Middleware`. As a result, the set of functions that can be considered `Middleware` is limited in scope by the arity of the function as well as the arguments and return types. It becomes easy to determine what functions implement `Middleware` in the absence of Actix-style attribute macros.

Both the implementer and the caller of `Finalize` can customize metadata such as the status code, headers, and cookies by using the `ResponseBuilder` API. This is contrary to the common design pattern that uses type-level composition with an extension trait that would encourage authors of middleware functions to return an unnameable type (via `impl Trait`) in order to elide the implementation details of how the response was built from the function signature. Type-level composition is great and an important part of being a Rust developer. However, Via prefers to stick with a remix of the builder pattern for the sake of generating responses from arbitrary types.


src/error/rescue.rs
```diff
@@ -130,7 +130,7 @@ impl Pipe for Sanitizer<'_> {
                 },
             );
 
-            Json(&payload).pipe(response.status(status_code))
+            Json(&payload).finalize(response.status(status_code))
```
zacharygolba: Delegation and branching feel less weird now. It also encourages the author to think about branching the way they would as if they were writing an extension trait for `ResponseBuilder`.

> Move as little as possible, hoist branching and pre conditions to the top of the finalize implementation.

examples/echo-server/src/main.rs
```diff
@@ -1,21 +1,17 @@
 use std::process::ExitCode;
-use via::{App, Error, Next, Pipe, Request, Response, Server};
+use via::response::{Finalize, Response};
+use via::{App, Error, Next, Request, Server};
 
 async fn echo(request: Request, _: Next) -> via::Result {
-    request.pipe(Response::build())
+    request.finalize(Response::build())
```
zacharygolba: Perhaps this could be a default impl of an `into_response()` fn in the `Finalize` trait?

zacharygolba: I'm going to add this in a separate PR. This makes a ton of sense.

src/request/request.rs
```diff
@@ -211,10 +210,11 @@ impl<State> Request<State> {
...
 }
 
-impl<State> Pipe for Request<State> {
-    fn pipe(self, builder: ResponseBuilder) -> Result<Response, Error> {
-        let response = match self.headers().get(CONTENT_LENGTH) {
-            Some(len) => builder.header(CONTENT_LENGTH, len),
+impl<State> Finalize for Request<State> {
+    #[inline]
+    fn finalize(self, builder: ResponseBuilder) -> Result<Response, Error> {
+        let response = match self.headers().get(CONTENT_LENGTH).cloned() {
```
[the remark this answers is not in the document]
zacharygolba: Also, it's nice to think about implementations of `finalize` like a constructor in that it is a place where allocations are expected. A similarity that wasn't immediately obvious until the name change.

PR #314, opened 2025-10-25
feat(via): yield once when at capacity

papa-zakari:
Prefer yielding once when at capacity rather than involving a timer. 

This change lowers the attack surface error of the accept loop. Timers add complexity regardless of how well they are implemented.

The most likely scenario is that a permit will be returned to the semaphore when the pending connection tasks are polled after yield now. If there a still no permits available after yielding once, drop the connection to indicate to upstream load balancers that it's time to try another node.

I have a pretty high degree of confidence that this change will slightly reduce latency when at capacity in addition to removing the possibility of side channel in the accept loop. Simply dropping the connection is also fine if you are running more than one node with a load balancer like NGINX, HAProxy, or a similar cloud offering such as AWS's network or application load balancer.

Yielding back to the runtime is an optimistic bet that polling pending connections before trying another node is lower latency than that of the private network with which the load balancer delegates to your app server. It also is an extremely lightweight form a back pressure that allows other nodes in the cluster to poll pending connections or accept a new one.

### DIFF
--- a/src/server/accept.rs
+++ b/src/server/accept.rs
@@ -73,7 +73,7 @@ where
     let mut connections = JoinSet::new();
 
     // Start accepting incoming connections.
-    let exit_code = 'accept: loop {
+    let exit_code = loop {
         let (tcp_stream, _) = tokio::select! {
             // A new TCP stream was accepted from the listener.
             result = listener.accept() => match result {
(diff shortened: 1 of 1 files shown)


src/server/accept.rs
```diff
@@ -98,25 +98,12 @@ where
...
-
-                            // The server is still at capacity. Reset the connection.
-                            Err(_) => continue,
-                        }
-                    }
-
-                    _ = shutdown_rx.changed() => {
-                        break receive_ctrl_c!(shutdown_rx);
-                    }
+                task::yield_now().await;
```
zacharygolba: The cost of yielding here is likely <10 µs for most applications.

We are betting that a scheduler tick is all we need to stabilize a saturated cluster.

Any more and it wouldn't be worth it. Any less is no better than the status quo.

PR #327, opened 2025-10-30
refactor(via): correctness improvements to the websocket API

papa-zakari:
Correctness improvements to the websocket API. Stay tuned for a more in-depth description.

### DIFF
--- a/src/payload.rs
+++ b/src/payload.rs
@@ -128,10 +128,11 @@ impl Payload for Bytes {
         dest
     }
 
-    fn serde_json_untagged<T>(self) -> Result<T, Error>
+    fn serde_json_untagged<T>(mut self) -> Result<T, Error>
     where
         T: DeserializeOwned,
     {
-        deserialize_json(self.as_ref())
+        let detached = self.split_to(self.len());
+        deserialize_json(detached.as_ref())
     }
 }


via-router/src/router.rs
```diff
@@ -8,7 +10,10 @@ use crate::path::{self, Param, Pattern, Split};
 ///
 pub struct Route<'a, T>(MatchCond<slice::Iter<'a, MatchCond<T>>>);
 
-pub struct RouteMut<'a, T>(&'a mut Node<T>);
+pub struct RouteMut<'a, T> {
+    node: &'a mut Node<T>,
+    _rc: PhantomData<Rc<()>>,
```
zacharygolba: This field makes `RouteMut: !Send + !Sync`. Preventing it from being moved into a route handler.

examples/chat/src/main.rs
```diff
@@ -1,37 +1,88 @@
...
-        room.route("/join").to(via::ws(room::join));
+        let (collection, member) = via::rest!(routes::users);
 
-        // show the message with the provided index.
-        room.route("/:index").to(via::get(room::show));
-    }
+        users.route("/").to(collection);
+        users.route("/:id").to(member);
+        users.route("/login").to(via::post(login));
+        users.route("/logout").to(via::post(logout));
```
zacharygolba: These are the only 2 routes that require an allocation to resolve. It's because there is a dynamic pattern adjacent to sibling routes with a static pattern.

From a router allocation perspective, this gives REST the same routing allocation behavior as GraphQL.

examples/chat/src/main.rs
```diff
@@ -1,37 +1,88 @@
...
+    api.route("/threads").scope(|threads| {
+        let mut thread = {
+            let (collection, member) = via::rest!(routes::threads);
+
+            threads.route("/").to(collection);
+            threads.route("/:thread-id").to(member)
+        };
+
+        // 403 when the current user is not in the thread.
+        thread.uses(routes::threads::authorization);
```
zacharygolba: We'll probably attempt to statically determine if a user is participating in a thread. Perhaps we'll have a list of threads to which they belong in the cookie or JWT. This doesn't scale to an unbounded number of threads but it's probably fine for this example. This is pretty robust as it is.

It's possible that we'll defer authorization until the database query is made and confirm that a join table exists with the user_id and thread_id as a part of the database query and use the `FoundOrForbidden` trait to propagate authorization errors like we do for messages and reactions.

examples/chat/src/util/auth.rs
```diff
@@ -0,0 +1,67 @@
...
+            && let Ok(user) = serde_json::from_str(cookie.value())
+        {
+            request.extensions_mut().insert(Verify(user));
+        }
+
+        next.call(request)
+    }
+}
+
+impl Authenticate for Request {
```
zacharygolba: I'm considering making request a container type. Users would have to either take a reference to the request head or move the request head and body out of request before accessing the fields they want to access.

It sounds cumbersome but the explicit moves in the route modules are really nice to read and work with.

examples/chat/src/routes/messages.rs
```diff
@@ -0,0 +1,164 @@
...
+    if cfg!(debug_assertions) {
+        println!("\n{}", diesel::debug_query::<Pg, _>(&delete));
+    }
+
+    // Acquire a database connection and execute the delete.
+    {
+        let pool = request.state().pool();
+        let mut conn = pool.get().await?;
+
+        delete.execute(&mut conn).await.found_or_forbidden()?;
```
zacharygolba: I really like this pattern for access control. It doesn't fail early and allows for human error but it works exactly as you would expect without any extra database queries. A good reason for logging the query in debug mode.

examples/chat/src/models/message.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    pub author_id: Option<Uuid>,
+    pub thread_id: Option<Uuid>,
+}
+
+#[derive(Serialize)]
+pub struct MessageWithJoins {
+    #[serde(flatten)]
+    message: Message,
+    thread: Thread,
```
zacharygolba: I'm not going to include thread by default in the message joins. The way that the urls are nested implies that the user already has the thread loaded in to memory. I'll definitely load the first 10 or so reactions as a part of the query as well as the total number of reactions - the amount included in the message payload to show the amount of reactions that are below the fold.

examples/chat/src/models/message.rs
```diff
@@ -0,0 +1,113 @@
...
+    author: User,
+}
+
+pub fn by_author(id: Uuid) -> Eq<col::author_id, Uuid> {
+    col::author_id.eq(id)
+}
+
+pub fn by_cursor(
+    cursor: (NaiveDateTime, Uuid),
+) -> And<Eq<col::created_at, NaiveDateTime>, Lt<col::id, Uuid>, Bool> {
```
zacharygolba: These return types are a bit gnarly but it beats loading the schema DSL for the table in the routes module and turning your editor into a big ol' bag of skittles.

They look like named rails scopes on a model but they are composable and do not execute queries. It's a pretty cool pattern and makes me happy that I decided to use an ORM.

examples/chat/src/routes/users.rs
```diff
@@ -0,0 +1,67 @@
...
+    Ok(response)
+}
+
+pub async fn login(request: Request, _: Next) -> via::Result {
+    let (head, future) = request.into_future();
+    let params = future.await?.serde_json::<LoginParams>()?;
+    let user = User::query()
+        .filter(by_username(&params.username))
+        .first(&mut head.state().pool().get().await?)
+        .await?;
```
zacharygolba: Leaving the door open for now for a password free option that works out of the box. Worst case scenario, I'll save passwords in the database and put comments above every line of code that references a password reminding that an alternative should be preferred in production.

examples/chat/src/util/auth.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait Authenticate {
+    fn current_user(&self) -> via::Result<&User>;
+}
+
+pub struct Auth {
+    cookie: String,
+}
+
+#[derive(Clone)]
+struct Verify(User);
```
zacharygolba: This prevents malicious middlewares from accessing the `CurrentUser` stored in the request extensions.

Hyper does the same thing for websocket upgrades and should be suitable for any auth plugin someone may be considering writing.

I can't imagine any other option being more secure. The type visibility rules of Rust prevent this type from being spoofed. We believe the borrow checker when it gives us compiler errors related to memory safety so we should do the same and accept that this type cannot be named outside of this module.

src/ws/message.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub async fn send(&mut self, message: impl Into<Message>) -> super::Result<()> {
+        if self.0.send(message.into()).await.is_err() {
+            Err(ControlFlow::Break(ErrorKind::CLOSED.into()))
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn recv(&mut self) -> impl Future<Output = Option<Message>> {
+        coop::unconstrained(self.1.recv())
```
zacharygolba: Both of the mpsc receivers are unconstrained. This keeps the channel latency near-zero.

The inner loop of the start fn in the ws::upgrade module keeps the websocket handler cooperative while allowing a buffer size of 1. The channel is more like a callback than a form of back pressure.

This should scale nicely. It keeps the user code removed from the actual socket without the latency and queuing associated with a channel.

src/ws/upgrade.rs
```diff
@@ -0,0 +1,270 @@
...
+                    if let Err(error) = result {
+                        try_rescue_ws(error)
+                    } else {
+                        Continue(None)
+                    }
+                }
+
+                // Forward the incoming message to the channel.
+                Some(result) = stream.next() => {
+                    coop::consume_budget().await;
```
zacharygolba: This provides a bit of cushion by leaving a message on the stack if reading the next value out of the stream exceeds the budget for the task.

We may have to manually poll the listener future in order to ensure that the subsequent call to `tx.send` doesn't result in a deadlock but I'm pretty sure that yielding to the runtime is the right call here since the receiving half of the channel is unconstrained.

I may pull some people from the tokio group on discord in to the conversation here to confirm that this is the case. Otherwise, this will need a `poll_fn` to poll the listener on the subsequent line. In which case, I'd move the listener arm to the bottom of the select macro.

examples/chat/Cargo.toml
```diff
@@ -4,12 +4,20 @@ version = "0.1.0"
...
+chrono = { version = "0.4", features = ["serde"] }
+cookie = { version = "0.18", features = ["signed", "private"] }
+diesel = { features = ["chrono", "uuid"], version = "2" }
+diesel-async = { version = "0.7", features = ["bb8", "postgres"] }
+dotenvy = "0.15"
 http = "1"
 serde = "1"
 serde_json = "1"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+uuid = { version = "1", features = ["v4", "serde"] }
```
zacharygolba: I'm going to implement a hash id wrapper on top of uuid so they don't leak into urls. There will be a FromStr and Deserialize impl to make the hashing opaque for everyone.

examples/chat/src/routes/messages.rs
```diff
@@ -71,11 +72,13 @@ pub async fn create(request: Request, _: Next) -> via::Result {
...
-    head.state()
-        .publish(current_user_id, thread_id, Event::Message(&message))?;
+    let context = EventContext::new(Some(thread_id), current_user_id);
+    let event = Event::Message(message);
 
-    Response::build().status(201).json(&message)
+    // Notify subscribers that a message has been created and respond.
+    head.state()
+        .publish(context, event)?
+        .finalize(Response::build().status(201))
```
zacharygolba: I'm really happy with how the `Finalize` trait composes with a `ResponseBuilder`. This solves a rather complex problem in a pretty simple way. A trend that we want to continue with this framework.

PR #375, opened 2026-02-09
feat(via): simplify Payload trait with buf zeroing fns

papa-zakari:
This PR introduces a high-assurance mechanism for handling payload buffers in our web framework. It adds best-effort, opt-in zeroization for uniquely-owned Bytes frames and enforces defensive poisoning to make double-read attempts noisy, aligning with high-security practices suitable for both financial and military contexts.

The change aligns the framework with defense-grade memory handling practices, making payload processing robust against both accidental leaks and deliberate misuse. The opt-in zeroization gives application developers a choice: a bank’s engineering team may be reluctant to use APIs that deliberately zero buffers, as an overabundance of caution to avoid raising concerns with a compliance team, whereas a military-focused team might prefer a framework that provides zeroing by default to protect sensitive data.

These safeguards are essential when implementing APIs that handle secret or sensitive values at scale — for example, a service similar to AWS Systems Manager Parameter Store, where encryption, controlled memory handling, and strict read-once semantics are critical to protecting configuration secrets and sensitive parameters in multi-tenant environments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,8 +96,8 @@ optional = true
 default-features = false
 
 [dependencies.tokio-websockets]
-version = "0.12"
-features = ["aws_lc_rs", "server"]
+version = "0.13"
+features = ["server"]
 optional = true
 
 [dev-dependencies]


src/request/payload.rs
```diff
@@ -183,75 +173,43 @@ impl Aggregate {
...
-impl Payload for Bytes {
-    fn coalesce(mut self) -> Result<Vec<u8>, Error> {
-        let data = self.to_vec();
-        self.advance(self.len());
-        Ok(data)
-    }
+        for frame in self.payload.iter_mut() {
+            let ptr = frame.as_ptr() as *mut u8;
+            let len = frame.len();
+            let src = unsafe { slice::from_raw_parts_mut(ptr, len) };
```
zacharygolba: This commit needs cleaned up a bit.

We still need safety comments and a review from another human.

PR #380, opened 2026-02-20
fix(via): simplified examples in workspace

papa-zakari:
A work in progress effort of moving the more complicated examples to a separate repo within the Via org.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ optional = true
 default-features = false
 
 [dev-dependencies]
-cookie = { version = "0.18", features = ["percent-encode", "private"] }
+cookie = { version = "0.18", features = ["percent-encode", "signed"] }
 rustls-pemfile = "2"
 time = { version = "0.3" }
 tokio = { version = "1", features = ["macros", "io-util", "rt-multi-thread"] }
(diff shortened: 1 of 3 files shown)


src/app/shared.rs
```diff
@@ -242,83 +230,104 @@ use std::sync::Arc;
...
-///
-///     Response::build()
-///         .status(StatusCode::NO_CONTENT)
-///         .finish()
-/// }
-/// ```
+///     // This example favors anonymity over performance. Therefore, we clone
+///     // app as early as we can.
+///     //
+///     // Any earlier and Uuid parse errors could become a DoS vector that
```
zacharygolba: While it would be very difficult (albeit nearly impossible) to exploit an Arc as a DoS vector, contention in general implies that deadlocks are possible and I don't know how else I'd classify this type of risk.

The idea is to keep contention low by having atomics follow a predictable rhythm. The RwLock is far more likely to deadlock (what's in a name).

However, increasing contention on a `Shared<App>` could increase the window of opportunity that an attack could have to steal a buffer that is waiting to be dropped or zeroed.

I like being specific here but IMO, treating Arc as if it can "soft deadlock" is the safest approach and easiest way to teach this concept.

Can't wait for community input! It sounds sarcastic but I'm so sincere.

PR #383, opened 2026-02-25
refactor(via-router): cleanup traversal algorithm

papa-zakari:
A work-in-progress to effort to clean up the router traversal algorithm. There are probably a couple misnomers and name changes that have to happen before this gets merged. The functionality should be identical and it should be as safe if not more safe than before.

Rust makes the memory safety aspect easy.

The rest is just theoretical correctness. The biggest challenge will be not letting the theoretical correctness get in the way of practical correctness.

Perhaps fully lazy iteration is a better approach but eagerly traversing all nodes at a given depth may be beneficial considering that the `Traverse` struct iterator implementation is intended to be fully consumed all the time.

It terminates early for the sake of safety anyway.

### DIFF
--- a/via-router/src/path.rs
+++ b/via-router/src/path.rs
@@ -117,33 +117,25 @@ impl<'a> Iterator for Split<'a> {
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        let offset = &mut self.offset;
-        let start = *offset;
-        let path = self.path;
+        let start = self.offset;
 
-        if let Some(len) = path
-            .get(start..)?
+        if start == self.path.len() {
+            return None;
+        }
+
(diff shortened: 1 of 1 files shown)


via-router/src/router.rs
```diff
@@ -36,7 +36,6 @@ enum MatchCond<T> {
 /// A group of nodes that match the path segment at `self.range`.
 ///
 struct Binding<'a, T> {
-    exhausted: bool,
     results: SmallVec<[&'a Node<T>; 1]>,
```
zacharygolba: In practice, the results vec len is almost always 1. There is usually a better way to structure the route tree in cases where the smallvec would spill onto the heap.

PR #390, opened 2026-03-03
feat(via): uniform allocation sizes

papa-zakari:
This change standardizes internal memory allocations across the request processing pipeline to reduce side-channel leakage and improve determinism for high-assurance use cases.

Key changes:

- Side-channel mitigation: identical allocation sizes prevent observers from inferring payload size, path complexity, or middleware depth.

- Deterministic memory layout: fixed-capacity vectors avoid unexpected heap growth during request processing.

- Reduced panic leakage risk: preallocated frames reduce the probability of exposing a large contiguous buffer on task panic.

- Simplicity and auditability: memory usage is consistent and easy to reason about, supporting high-assurance / security-critical environments.


src/request/params.rs
```diff
@@ -7,15 +7,15 @@ use super::query::QueryParser;
 use crate::util::UriEncoding;
 use crate::{Error, raise};
 
-pub(super) type ParamRange = (usize, Option<usize>);
+pub(super) type ParamRange = [Option<usize>; 2];
```
zacharygolba: We're trading a branch that will basically never run for noise. It's the right decision, IMO.

PR #398, opened 2026-03-04
fix(via): select connection version based on alpn version

papa-zakari:
Avoids HTTP version based feature flags in favor of automatically selected the http version based on alpn negotiation.

This PR introduces a feature that is _essential_ for the seamless use of Via in production.

---

_Note:_ tokio-native-tls does not seem to handle alpn negotation directly.

I'm personally unaware of how one configures alpn protocols for the various native TLS implementations on each platform.

However, if any work has to be done on the tokio-native-tls side or if possible—configuring a system-wide preferred alpn protocol order for the native platform implementations should work with our `NativeTlsAcceptor` with no additional changes to our code.

### DIFF
--- a/examples/tls/rustls.rs
+++ b/examples/tls/rustls.rs
@@ -50,7 +50,7 @@ mod tls {
             .with_single_cert(certs, key)
             .expect("tls config is invalid or missing");
 
-        config.alpn_protocols = vec![b"h2".to_vec()];
+        config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
 
         config
     }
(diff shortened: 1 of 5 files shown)


examples/tls/rustls.rs
```diff
@@ -50,7 +50,7 @@ mod tls {
             .with_single_cert(certs, key)
             .expect("tls config is invalid or missing");
 
-        config.alpn_protocols = vec![b"h2".to_vec()];
+        config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
```
zacharygolba: Users of Via that enable the rustls feature flag can specify the order of http versions that they prefer their clients to use.

PR #414, opened 2026-03-09
feat(via): improve cse of mask for http method dispatch

papa-zakari:
Improves the probability of `mask` being reduced as a CSE while dispatching middleware based on the HTTP method. This is achieved by:

- Removing the intermediate ref to `method` from which mask is computed.
- Changing the Predicate trait to take the rhs of the operation by reference.

### DIFF
--- a/src/router/allow.rs
+++ b/src/router/allow.rs
@@ -29,7 +29,7 @@ pub(crate) struct MethodNotAllowed {
 }
 
 trait Predicate {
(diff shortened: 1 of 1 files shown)


zacharygolba:
From a performance perspective this is a great sign. From a security standpoint, it's even better. This is likely compiling to the equivalent `match method` where method is an `i8`. :shipit:

PR #465, opened 2026-03-30
feat(examples): add an in-memory semi-ephemeral chat example

papa-zakari:
A work-in-progress implementation of redis pubsub in a chat app backed by postgres and diesel.

The chat flow is working with via-pubsub If you want to test it, checkout this branch and run the following:

```
cd examples/chat
docker-compose up -d
cp example.env .env
cargo run --example chat --features="aws-lc-rs tokio-tungstenite"
```


examples/chat/src/main.rs
```diff
@@ -53,22 +53,29 @@ async fn main() -> Result<ExitCode, Error> {
     // Parse and track changes that are made to the session cookie.
     api.middleware(cookies([session::COOKIE]));
 
-    // Confirm that the client speaks JSON, then load the session from the
-    // session cookie if it is present in the request.
+    // Content negotiation and authentication guards.
     api.middleware(guard::flat_map(
```
papa-zakari, line 65: This middleware composition demonstrates the power of the latest guard API.

examples/chat/src/main.rs
```diff
@@ -53,22 +53,29 @@ async fn main() -> Result<ExitCode, Error> {
...
+        middleware(|mut request, next| {
+            // If the request is read only or the session was verified in the
+            // past hour, skip verifying that the user still has an account.
+            //
+            // This is safe so long as your app is the authoritative source of
+            // truth for the session and you properly end websocket sessions if
+            // the user deletes their account.
+            let auth = guard::filter(
+                guard::or((method::is_mutation(), session::is_stale())),
+                session::refresh(),
```
papa-zakari: Either the future returned by `session::refresh` or calling `next` is when async work begins.

examples/chat/src/routes/users.rs
```diff
@@ -0,0 +1,184 @@
...
+        deny!(403, "only the account owner can update a user");
+    }
+
+    // Prepare to read the request body.
+    let (body, app) = request.into_future();
+
+    // Apply the change set to the active user.
+    let user = {
+        // Acquire a database connection and read the request body.
+        let (mut conn, payload) = tokio::try_join!(
```
papa-zakari: This is a great optimization but likely compromises the security of action a bit.

It's probably best if we wait until the last minute before acquiring a database connection.

We can be sure that reading a request body is secure because of the invariants upheld by the API contract of Payload. If you see a bug, please go through the proper channels for responsible disclosure.

Therefore, we should favor mitigating DoS risk by acquiring the connection after validating the contents of the request body.

papa-zakari:
As soon as tests pass, this can merge. We can finish the rest of the chat example incrementally. If we keep chat in a separate branch, drift is more likely and conflicts are a pain.

examples/chat/src/models/user.rs
```diff
@@ -0,0 +1,291 @@
...
+    #[serde(with = "time::serde::rfc3339")]
+    updated_at: OffsetDateTime,
+}
+
+#[derive(Deserialize)]
+pub struct NewUser {
+    email: String,
+    username: String,
+    #[serde(rename = "password")]
+    password_hash: Password,
```
papa-zakari, line 43: We may have to consider a two-phased deserialization step for new users in order to compare passwords with password confirmations.

We can revisit this when we implement a reset password flow.

papa-zakari:
I'm going to go ahead and merge this. We can continue to make progress on this incrementally as a pre-requisite of the official 2.0 release.

PR #512, opened 2026-04-12
feat(via): mod guards exposes a combinator api

papa-zakari:
A reworked guard API complete with docs that _should_ compiles down to C-style boolean logic gates.

```rust
use via::guard::{self, header::media};

// If the client does not speak JSON, deny the request.
app.middleware(guard::content(media::json(), media::json()));
```

**Slightly Desugared:**

```rust
use via::guard::{guard, header, method, when};

// If the client does not speak JSON, deny the request.
app.middleware(guard((
    header::accept(header::media::json()),
    when(
        method::is_mutation(),
        (
            header::content_type(header::media::json()),
            header::content_length(),
        ),
    ),
)));
```


papa-zakari:
Error is definitely going to be a associated type on Predicate.

Byte level combinators can use a ZST for the Err discriminant so they become a bool.

PR #514, opened 2026-04-17
fix(via): error response uses a json array for errors

papa-zakari:
Fixes a regression where errors in error responses were conditionally serialized as a JSON array.

### DIFF
--- a/via/src/error/mod.rs
+++ b/via/src/error/mod.rs
@@ -7,7 +7,10 @@ mod result;
 mod server;
 
 use http::{StatusCode, header};
+use serde::ser::SerializeSeq;
 use serde::{Serialize, Serializer};
+use smallvec::SmallVec;
+use std::borrow::Cow;
 use std::fmt::{self, Debug, Display, Formatter};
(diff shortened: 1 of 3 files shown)


via/src/error/mod.rs
```diff
@@ -204,23 +202,31 @@ impl Error {
...
-            Errors::new(self.status, message)
-        } else {
-            let mut errors = Vec::with_capacity(12);
-            let mut source = self.source();
-
-            while let Some(error) = source {
-                errors.push(error.to_string());
-                source = error.source();
-            }
+    fn as_message_or_source(&self) -> Result<&str, Option<&(dyn std::error::Error + 'static)>> {
```
papa-zakari, line 205: This seems a bit excessive but it's exactly what we want. We either have an error message that we can take a reference to and serialize without an allocation or an error chain with a length of at least 1. Eagerly upcasting to a `dyn Error + 'static` allows us to use the same reference that we match on for the optimized case while only introducing 1 extra branch as opposed to a branch per source.

PR #520, opened 2026-04-19
feat(via): timeout helpers for reading a request body

papa-zakari:
Provides an easy way for enforcing a timeout when reading a request body.

We want to make the process of adding a timeout to reading a request body as seamless as possible so DoS vectors that involve resource retention (in this case, heap memory + fd) become, for the most—inexcusable.

This is implemented internally as a macro rather than a trait because applying a timeout is something that can be done to any Future and we only have two types that we want to implement these helper functions for. Sealing the trait is excessive and opening it is confusing because it creates an alternative syntax to the timeout macro that really only makes sense for this use-case.

### DIFF
--- a/examples/json.rs
+++ b/examples/json.rs
@@ -23,7 +23,7 @@ async fn hello(request: Request, _: Next) -> via::Result {
     let (body, _app) = request.into_future();
 
     // Aggregate the frames of the request body and then deserialize a Document<Hello>.
-    let hello: Document<Hello> = body.await?.json()?;
+    let hello: Document<Hello> = body.timeout_after_secs(5).await?.json()?;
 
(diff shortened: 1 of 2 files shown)


examples/json.rs
```diff
@@ -23,7 +23,7 @@ async fn hello(request: Request, _: Next) -> via::Result {
     let (body, _app) = request.into_future();
 
     // Aggregate the frames of the request body and then deserialize a Document<Hello>.
-    let hello: Document<Hello> = body.await?.json()?;
+    let hello: Document<Hello> = body.timeout_after_secs(5).await?.json()?;
```
papa-zakari, line 26: We want the barrier to be this low for adding a timeout.

PR #532, opened 2026-04-28
refactor(via): ws upgrade flow is as clean as possible

papa-zakari:
Refactors the websocket upgrade flow to be as clean as possible (at times, paradoxically). This PR does not change the functionality of the websocket upgrade in any way.

Instead, it does the following:

- Sequences upgrade steps to avoid creating signals through explicit state transitions that can be observed outside the application
  - Prefers a uniform allocation size instead of different sized allocations that indicate an error occurring during a validation step
  - Defers different sized allocations that occur for errors vs. responses into the returned BoxFuture to utilize the natural timing entropy from the runtime 

- Obfuscates critical upgrade steps with noisy or entropic operations 
  - The request envelope arc allocation occurs after the response is built rather than before (allocator entropy makes timing non deterministic)
  - The actual upgrade handshake occurs adjacent to a bunch of memcpys drowning out any signal that comes from a move with noise

- Makes the spawned reactor task an async function to make it obvious to the reader what is moved into the spawned task

- Enforces parity between the configuration used by both ws backends
  - `buffer_size == frame_size == max_message_size` limits the number of I/O syscalls per wake to be the amount required to receive, send, and flush a single message while preferring the smallest reasonable buffer size to limit the severity of a leak if the worst case scenario occurs

- Makes the actual upgrade handshake future unconstrained to prevent a connection that is ready for upgrade from yielding to the runtime

### DIFF
--- a/via/src/ws/request.rs
+++ b/via/src/ws/request.rs
@@ -1,3 +1,4 @@
+use hyper::upgrade::OnUpgrade;
 use std::sync::Arc;
 
 use crate::app::Shared;
@@ -5,6 +6,7 @@ use crate::request::Envelope;
 
 #[derive(Debug)]
 pub struct Request<App = ()> {
+    pub(super) on_upgrade: Option<OnUpgrade>,
     envelope: Arc<Envelope>,
     app: Shared<App>,
 }
@@ -25,9 +27,10 @@ impl<App> Request<App> {
 
 impl<App> Request<App> {
     pub(super) fn new(request: crate::Request<App>) -> Self {
-        let (envelope, _, app) = request.into_parts();
+        let (mut envelope, _, app) = request.into_parts();
 
         Self {
+            on_upgrade: envelope.extensions_mut().remove(),
             envelope: Arc::new(envelope),
             app,
         }
@@ -37,6 +40,7 @@ impl<App> Request<App> {
 impl<App> Clone for Request<App> {
     fn clone(&self) -> Self {
         Self {
+            on_upgrade: None,
             envelope: Arc::clone(&self.envelope),
             app: self.app.clone(),
         }
--- a/via/src/ws/upgrade.rs
+++ b/via/src/ws/upgrade.rs
@@ -1,29 +1,34 @@
 use http::StatusCode;
 use http::header::{self as h, HeaderMap};
-use hyper::upgrade::OnUpgrade;
 use std::future::Future;
 use std::sync::Arc;
 
 #[cfg(feature = "tokio-tungstenite")]
-use tokio_tungstenite::WebSocketStream;
+use tokio_tungstenite::{
+    WebSocketStream,
+    tungstenite::protocol::{Role, WebSocketConfig},
+};
 
 #[cfg(feature = "tokio-websockets")]
 use tokio_websockets::WebSocketStream;
 
-use super::Channel;
 use super::io::UpgradedIo;
 use super::run::RunTask;
 use super::util::{Base64EncodedDigest, sha1};
-use crate::guard::header::{CaseSensitive, Contains, Header};
-use crate::guard::{Predicate, header};
+use super::{Channel, Request};
+use crate::guard::header::{CaseSensitive, Contains, DenyHeader, Header};
+use crate::guard::{self, MapErr, Predicate, header};
 use crate::ws::error::UpgradeError;
-use crate::{BoxFuture, Error, Middleware, Next, Request, Response, ResultExt, deny};
+use crate::{BoxFuture, Error, Middleware, Next, Response, ResultExt};
 
 const DEFAULT_FRAME_SIZE: usize = 16384; // 16KB
 
 pub struct Ws<T> {
     listener: Arc<Listener<T>>,
-    guard: (Header<CaseSensitive>, Header<Contains>, Header<Contains>),
+    guard: MapErr<
+        fn(DenyHeader<'_>) -> UpgradeError,
+        (Header<CaseSensitive>, Header<Contains>, Header<Contains>),
+    >,
 }
 
 pub(super) struct Listener<T> {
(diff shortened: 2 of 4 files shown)


via/src/ws/request.rs
```diff
@@ -25,9 +27,10 @@ impl<App> Request<App> {
 
 impl<App> Request<App> {
     pub(super) fn new(request: crate::Request<App>) -> Self {
-        let (envelope, _, app) = request.into_parts();
+        let (mut envelope, _, app) = request.into_parts();
 
         Self {
+            on_upgrade: envelope.extensions_mut().remove(),
```
papa-zakari, line 33: This is the safest thing that we can do aside from leaving the upgrade callback in extensions.

If hyper supported implementing `CanUpgrade` for request structs with a custom layout, we'd use their API.

The only change I'm going to make is panicking if the callback is None. This prevents a ghost from eating the socket prior to the upgrade handshake completing.

They would presumably use the same task, if they don't it should be observable with tracing.

via/src/ws/upgrade.rs
```diff
@@ -1,108 +1,128 @@
...
+                    buffer_size: DEFAULT_FRAME_SIZE,
+                    max_frame_size: Some(DEFAULT_FRAME_SIZE),
+                    max_message_size: Some(DEFAULT_FRAME_SIZE),
+                },
             }),
-            guard: (
-                header(h::SEC_WEBSOCKET_VERSION, header::case_sensitive(b"13")),
-                header(h::CONNECTION, header::contains(header::tag(b"upgrade"))),
-                header(h::UPGRADE, header::contains(header::tag(b"websocket"))),
+            guard: guard::map_err(
```
papa-zakari: Moving the MapErr step to the pre-composed guard defines a structure that is easier for rustc to optimize. `map_err` enforces that the closure implement `Copy` preventing the environment from being captured by this closure.

via/src/ws/upgrade.rs
```diff
@@ -138,68 +158,39 @@ impl<T> Ws<T> {
...
-                };
-
-                if let Err(error) = task.await {
-                    log!(error(0), "{}", &error);
-                }
-            });
-
-            Response::build()
+        Box::pin(async {
+            let accept = is_valid.or_bad_request()?;
```
papa-zakari, line 179: Allocating for the error type is deferred so it's impossible to predict the outcome of the upgrade until we are ready to generate the response.

via/src/ws/upgrade.rs
```diff
@@ -138,68 +158,39 @@ impl<T> Ws<T> {
...
+                .finish();
 
-impl Default for WsConfig {
-    fn default() -> Self {
-        Self {
-            buffer_size: DEFAULT_FRAME_SIZE,
-            max_frame_size: Some(DEFAULT_FRAME_SIZE),
-            max_message_size: Some(DEFAULT_FRAME_SIZE),
-        }
+            tokio::spawn(reactor(Request::new(request), listener));
```
papa-zakari, line 187: Allocating for the request envelope occurs after a number of allocations that occur when generating a response.

This adds timing entropy to the allocation for the envelope, making it harder to predict and less of a signal for when the reactor task is spawned.

via/src/ws/upgrade.rs
```diff
@@ -89,7 +90,7 @@ where
     T: Fn(Channel, Request<App>) -> Await + Send,
     Await: Future<Output = super::Result> + Send + 'static,
 {
-    let err = match handshake(&mut request, &listener.config).await {
+    let err = match unconstrained(handshake(&mut request, &listener.config)).await {
```
papa-zakari, line 90: The upgrade handshake should only yield if the upgrade response has not been written to the socket.

via/src/ws/upgrade.rs
```diff
@@ -117,13 +114,10 @@ impl<T> Ws<T> {
...
                 },
             }),
-            guard: guard::map_err(
-                |error| error.into(),
-                (
-                    header(h::SEC_WEBSOCKET_VERSION, header::case_sensitive(b"13")),
-                    header(h::CONNECTION, header::contains(header::tag(b"upgrade"))),
-                    header(h::UPGRADE, header::contains(header::tag(b"websocket"))),
-                ),
+            guard: (
```
papa-zakari, line 117: `guard::MapErr` is perfectly secure but deferring the error conversion ensures that the closure is unique with each invocation of `Ws::validate` and that the closure type is unnamable.

via/src/ws/upgrade.rs
```diff
@@ -161,12 +155,13 @@ impl<T> Ws<T> {
 impl<T> Ws<T> {
     #[inline]
     fn validate(&self, headers: &HeaderMap) -> Result<Base64EncodedDigest, UpgradeError> {
-        self.guard
-            .cmp(headers)
-            .and_then(|_| match headers.get(h::SEC_WEBSOCKET_KEY) {
+        self.guard.cmp(headers).map_or_else(
```
papa-zakari, line 158: I draw the line with my paranoia where `rustc` would not generate the same instructions (if not better) with a nested match.

It's reasonable to think that an `.so` can be loaded into memory and it contains a fn pointer that fits perfectly in the slot where the fn pointer for our pre-built map_err guard lived. Even if it's in `.rodata`. We can't assume our users use immutable deploys in a k8s cluster with a readonly image that replaces a node if the process exits. Our users could very well be using `systemd`.

That's why we call ourselves high assurance.

It's unreasonable for `rustc` to generate code that is less secure if you use `map_or_else` instead of a nested match when the implementation is literally:

```rust
#[inline]
#[stable(feature = "result_map_or_else", since = "1.41.0")]
#[rustc_const_unstable(feature = "const_result_trait_fn", issue = "144211")]
pub const fn map_or_else<U, D, F>(self, default: D, f: F) -> U
where
    D: [const] FnOnce(E) -> U + [const] Destruct,
    F: [const] FnOnce(T) -> U + [const] Destruct,
{
    match self {
        Ok(t) => f(t),
        Err(e) => default(e),
    }
}
```

via/src/ws/upgrade.rs
```diff
@@ -31,73 +34,85 @@ pub(super) struct Listener<T> {
...
+async fn handshake<App>(
+    request: &mut Request<App>,
     config: &WsConfig,
-    upgrade: OnUpgrade,
 ) -> Result<WebSocketStream<UpgradedIo>, Error> {
-    use tokio_websockets::server::Builder;
-    use tokio_websockets::{Config, Limits};
+    use tokio_websockets::{Config, Limits, server::Builder};
 
+    let on_upgrade = request.on_upgrade.take().expect("already upgraded.");
```
papa-zakari, line 75: If this value is `None`, it's safe to assume that the task is compromised. Panicking is the safest thing to do as it'll close the connection.

PR #534, opened 2026-04-28
fix(via): prefer an Arc::clone in rescue to guarantee size

papa-zakari:
Prefers the cost of an atomic operation in the `Rescue` middleware so the closure has a deterministic size. This ensures correctness for all cases of `F`.

### DIFF
--- a/via/src/error/rescue.rs
+++ b/via/src/error/rescue.rs
@@ -45,7 +45,7 @@ where
         Box::pin(async move {
             future.await.or_else(|mut error| {
                 let mut sanitizer = Sanitizer::new(&mut error);
-                (&*recover)(&mut sanitizer);
+                (recover)(&mut sanitizer);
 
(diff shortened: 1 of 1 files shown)


via/src/error/rescue.rs
```diff
@@ -38,13 +39,13 @@ where
...
     fn call(&self, request: Request<App>, next: Next<App>) -> BoxFuture {
+        let recover = Arc::clone(&self.recover);
         let future = next.call(request);
-        let recover = *self.recover;
 
         Box::pin(async move {
             future.await.or_else(|mut error| {
                 let mut sanitizer = Sanitizer::new(&mut error);
-                recover(&mut sanitizer);
+                (&*recover)(&mut sanitizer);
```
papa-zakari: It's safe to let this auto-deref. It's nice to see it explicit like this in the diff though.

PR #537, opened 2026-04-30
feat(via): ws io is statically dispatched

papa-zakari:
Improves the performance and reliability of the ws reactor by using a concrete Io type for `WebSocketStream<UpgradedIo>`.

This is accomplished by boxing the inner future returned by the `RustlsAcceptor`, ensuring that each TLS I/O object is allocated on the heap to reduce stack pressure of long lived connections.

This provides the same safety benefits as the previous implementation by using a boxed I/O type while maintaining a concrete type to reduce overhead of reads and writes on the I/O object.


via/src/server/accept.rs
```diff
@@ -25,6 +29,21 @@ macro_rules! log {
     };
 }
 
+macro_rules! serve_unless_cancelled {
+    ($cancellation:ident, $connection:ident) => {
```
papa-zakari, line 33: We use `ident` for both arguments. This allows us to shadow the previous binding and also guarantees that we're not illegally aliasing a pinned future or mutable ref to a future.

via/src/ws/io.rs
```diff
@@ -5,16 +5,23 @@ use std::pin::Pin;
...
 
 impl UpgradedIo {
     #[inline]
-    pub fn new(io: Upgraded) -> Self {
-        Self {
-            io: TokioIo::new(io),
-        }
+    pub fn new(io: Upgraded) -> Result<Self, UpgradeError> {
+        let Ok(parts) = io.downcast() else {
+            return Err(UpgradeError::Other);
```
papa-zakari, line 19: This is a pretty nice verification step that hardens the security of the websocket upgrade.

Dynamic dispatch theoretically allows spoofing. Albeit, it's unlikely.

We can safely assume that an application cannot continue to function as expected with a spoofed I/O struct that retains the type id and offset of the original `IoStream` type.

via/src/server/accept.rs
```diff
@@ -96,27 +111,34 @@ where
...
         #[cfg(any(feature = "native-tls", feature = "rustls-23"))]
         connections.spawn({
-            let handshake = timeout(
-                service.config().tls_handshake_timeout(),
-                acceptor.accept(io),
-            );
+            let handshake = acceptor.accept(io);
 
+            // native-tls task size: 1952
+            // rustls task size: 1712
```
papa-zakari, line 117: Anything less than `2048` should be sufficient to keep stack pressure low. Even for long-lived connections that use http2, a ws, or both.

User code sits behind a heap allocation, is type erased, and is dispatched dynamically.

This should guarantee stack stability, allowing the server to continue even if memory pressure is high. The most likely scenario of an OOM is a task panic, then either a quick recovery or a graceful shutdown starts due to an OOM error in the accept loop when attempting to accept the next connection.

This is enterprise-grade and high-assurance.

via/src/server/accept.rs
```diff
@@ -96,27 +111,34 @@ where
...
             async move {
-                let (io, alpn) = handshake.await??;
-                let io = IoWithPermit::new(io, permit);
+                let io = timeout(service.config().tls_handshake_timeout(), handshake).await??;
```
papa-zakari, line 119: 1 less memcpy that could be interpreted as signal if the timeout is moved into the spawned task.

via/src/server/accept.rs
```diff
@@ -216,19 +225,7 @@ where
...
-        biased;
-        // The connection future is ready.
-        result = connection.as_mut() => result?,
-        // A graceful shutdown signal was sent to the process.
-        _ = cancellation.wait() => {
-            connection.as_mut().graceful_shutdown();
-            connection.await?;
-        }
-    }
+    serve_unless_cancelled!(cancellation, connection);
```
papa-zakari, line 228: Using a macro prevents drift between http versions.

via/src/server/accept.rs
```diff
@@ -25,6 +29,21 @@ macro_rules! log {
     };
 }
 
+macro_rules! serve_unless_cancelled {
+    ($cancellation:ident, $connection:ident) => {
+        tokio::select! {
+            // The connection future is ready.
+            result = &mut $connection => result?,
```
papa-zakari, line 36: `tokio::select` is going to pin the future regardless of whether or not we give it `Pin<&mut impl Future>` so we might as well wait until the last minute.

via/src/server/tls/rustls.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let ReadyState::Handshake(accept) = &mut self.state else {
+            return Poll::Ready(Err(io::ErrorKind::AlreadyExists.into()));
+        };
+
+        Pin::new(accept).poll(cx).map_ok(|stream| {
+            let (_, conn) = stream.get_ref();
+            let alpn = match conn.alpn_protocol() {
+                Some(value) if value == b"h2" => Alpn::HTTP_2,
```
papa-zakari, line 119: `Alpn` can obfuscate the enum discriminant in a sea of bytes.

There's no reason not to use it instead of matching on the alpn bytes at the last minute.

PR #538, opened 2026-04-30
feat(via): http method mask is a u8 for obfuscation

papa-zakari:
This PR removes the connect method from http method based routing helpers. If you were using the `via::connect` fn, you can write a custom guard and flat_map it your middleware.

```rust
use http::Method;
use std::process::ExitCode;
use via::{Error, Next, Request, Server, guard};

#[tokio::main]
async fn main() -> Result<ExitCode, Error> {
    let mut app = via::app(());

    app.route("/connect").to(guard::flat_map(
        guard::map_err(
            |_| via::err!(405, "method not allowed."),
            |request: &Request| request.method() == Method::CONNECT,
        ),
        async |_: Request, _: Next| {
            todo!();
        },
    ));

    Server::new(app).listen(("127.0.0.1", 8080)).await
}
```

The reason why we are removing the `via::connect` fn is to obfuscate the signal that comes from a bitwise operation on a `u16`. In addition to `u8` being hard to recognize in a sea of bytes, it's also the size of the mask used in both uses of `tokio::select` in the accept loop and connection task respectively.

In addition to making the data type and value harder to recognize for the mask used to route dynamically dispatched middleware, this also makes it impossible to tell the difference between a GET or POST being routing to a middleware and polling a future in a `tokio::select` arm.

### DIFF
--- a/via/src/router/allow.rs
+++ b/via/src/router/allow.rs
@@ -31,16 +31,15 @@ pub(crate) struct MethodNotAllowed {
 
 bitflags! {
     #[derive(Clone, Copy, Debug, Eq, PartialEq)]
-    struct Mask: u16 {
-        const CONNECT = 1 << 0;
-        const DELETE  = 1 << 1;
-        const GET     = 1 << 2;
-        const HEAD    = 1 << 3;
-        const OPTIONS = 1 << 4;
-        const PATCH   = 1 << 5;
-        const POST    = 1 << 6;
-        const PUT     = 1 << 7;
-        const TRACE   = 1 << 8;
+    struct Mask: u8 {
+        const DELETE  = 1 << 0;
+        const GET     = 1 << 1;
+        const HEAD    = 1 << 2;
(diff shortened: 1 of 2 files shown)


via/src/router/allow.rs
```diff
@@ -31,16 +31,15 @@ pub(crate) struct MethodNotAllowed {
...
-        const GET     = 1 << 2;
-        const HEAD    = 1 << 3;
-        const OPTIONS = 1 << 4;
-        const PATCH   = 1 << 5;
-        const POST    = 1 << 6;
-        const PUT     = 1 << 7;
-        const TRACE   = 1 << 8;
+    struct Mask: u8 {
+        const GET     = 1 << 0;
+        const POST    = 1 << 1;
```
papa-zakari, line 36: The two most common HTTP methods can now be confused with the mask used to poll any of the futures in the tokio::select arms in accept.

PR #551, opened 2026-05-03
fix(via): async block wake semantics in via::ws

papa-zakari:
Fixes the ws listener to guarantee the same wake semantics as if it were implemented with an async block or async function. These wake semantics transcend to the listener. Meaning that you no longer have to await a future in order to uphold the reactor's invariants.

### DIFF
--- a/via/src/ws/channel.rs
+++ b/via/src/ws/channel.rs
@@ -26,6 +26,7 @@ struct Recv<'a> {
     recv: mpsc::Recv<'a, mpsc::Receiver<Message>>,
 }
 
+#[inline]
 fn poll_ready(tx: &mut Sender<Message>, cx: &mut Context) -> Poll<super::Result> {
     match tx.poll_ready(cx) {
         Poll::Pending => Poll::Pending,
@@ -98,9 +99,9 @@ impl Channel {
 impl Future for Recv<'_> {
     type Output = Option<Message>;
 
-    fn poll(mut self: Pin<&mut Self>, _: &mut Context) -> Poll<Self::Output> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         // Receive progress is driven by the I/O of `WebSocketStream`.
-        poll_immediate_no_wake(|noop| Pin::new(&mut self.recv).poll(noop)).map(Result::ok)
+        Pin::new(&mut self.recv).poll(cx).map(|result| result.ok())
     }
 }
 
--- a/via/src/ws/mod.rs
+++ b/via/src/ws/mod.rs
@@ -8,10 +8,10 @@ compile_error!("either \"aws-lc-rs\" or \"ring\" must be enabled to use the ws m
 compile_error!("features \"tokio-tungstenite\" and \"tokio-websockets\" are mutually exclusive.");
 
 macro_rules! log {
-    ($level:tt($indent:literal), $fmt:literal $($arg:tt)*) => {
+    ($level:tt($indent:expr), $fmt:literal $($arg:tt)*) => {
         #[cfg(debug_assertions)]
         eprintln!(
-            "{:indent$}{}(via::ws): {}",
+            "{:indent$}{}(ws): {}",
             "",
             stringify!($level),
             format_args!($fmt $($arg)*),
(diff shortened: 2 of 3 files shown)


via/src/ws/channel.rs
```diff
@@ -98,9 +99,9 @@ impl Channel {
 impl Future for Recv<'_> {
     type Output = Option<Message>;
 
-    fn poll(mut self: Pin<&mut Self>, _: &mut Context) -> Poll<Self::Output> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         // Receive progress is driven by the I/O of `WebSocketStream`.
-        poll_immediate_no_wake(|noop| Pin::new(&mut self.recv).poll(noop)).map(Result::ok)
+        Pin::new(&mut self.recv).poll(cx).map(|result| result.ok())
```
papa-zakari, line 104: Multiple wakes can be coalesced into a single wake. Debug logs show that a single wake occurs even if the receive half of the channel owned by the listener registers a wake.

via/src/ws/mod.rs
```diff
@@ -8,10 +8,10 @@ compile_error!("either \"aws-lc-rs\" or \"ring\" must be enabled to use the ws m
 compile_error!("features \"tokio-tungstenite\" and \"tokio-websockets\" are mutually exclusive.");
 
 macro_rules! log {
-    ($level:tt($indent:literal), $fmt:literal $($arg:tt)*) => {
+    ($level:tt($indent:expr), $fmt:literal $($arg:tt)*) => {
         #[cfg(debug_assertions)]
         eprintln!(
-            "{:indent$}{}(via::ws): {}",
+            "{:indent$}{}(ws): {}",
```
papa-zakari, line 14: You know the name of the framework already. This makes debug logs a bit less noisy.

via/src/ws/run.rs
```diff
@@ -93,6 +93,9 @@ impl Future for Facade {
     type Output = super::Result;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        #[cfg(debug_assertions)]
+        let mut i = 0;
```
papa-zakari, line 97: Indentation now indicates the number of state transitions per poll.

via/src/ws/run.rs
```diff
@@ -162,17 +163,24 @@ impl Future for Facade {
...
+                        log!(info(i + 2), "outbound message sent successfully.");
                         self.state = IoState::Receive;
+                        cx.waker().wake_by_ref();
                     } else {
-                        log!(info(2), "waiting for flush to complete.");
-                        return Poll::Pending;
+                        log!(info(i + 2), "waiting for flush to complete.");
                     }
+
+                    return Poll::Pending;
```
papa-zakari, line 176: Flush returns pending unconditionally. This is true even when a woken up in `IoState::Flush`. This guarantees stack pressure identical to that of an async function. The individual components move in lockstep. If backpressure forms in any layer it accumulates at the socket boundary. Exactly as it would if you were working directly with `WebSocketStream`.

via/src/ws/run.rs
```diff
@@ -162,17 +163,24 @@ impl Future for Facade {
...
 
                 IoState::Flush => {
-                    log!(info(0), "state = flush");
+                    log!(info(i), "state = flush");
 
                     if Pin::new(stream).poll_flush(cx).map_err(rescue)?.is_ready() {
-                        log!(info(2), "flush complete. transition to receive.");
+                        log!(info(i + 2), "outbound message sent successfully.");
                         self.state = IoState::Receive;
+                        cx.waker().wake_by_ref();
```
papa-zakari, line 171: Calling `wake_by_ref` is necessary after flush. The buffering defaults of the `WebSocketStream` configuration guarantee messages are flushed in a single wake cycle for the majority of messages.

Users can chose whether or not they prefer a larger buffer size over yielding to the runtime if woken up to complete a partial flush.

A staggering majority of use cases will never have to worry about this. If you notice flushes are interrupted by a yield to the runtime every so often, don't panic. This is a self-healing behavior that ensures consistency between each layer with the minimum amount of explicit buffering.

PR #555, opened 2026-05-11
feat(via): sec-websocket-accept is valid utf8

papa-zakari:
Reinforces the security of the ws upgrade handshake by validating the utf8-ness of the base64 encoded `sec-websocket-accept` header value.

This is probably unnecessary in most cases but it's such a low cost for added assurance that I wouldn't feel comfortable saying that we are a high assurance web framework without doing such a thing. Encoding also happens once per upgrade so the difference in performance is negligible.

### DIFF
--- a/via/src/ws/util/sha1.rs
+++ b/via/src/ws/util/sha1.rs
@@ -28,9 +28,8 @@ pub fn sha1(input: &[u8]) -> Result<Base64EncodedDigest, UpgradeError> {
 }
 
 impl Base64EncodedDigest {
-    #[inline(always)]
-    pub fn as_str(&self) -> &str {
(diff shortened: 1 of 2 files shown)


via/src/ws/util/sha1.rs
```diff
@@ -28,9 +28,8 @@ pub fn sha1(input: &[u8]) -> Result<Base64EncodedDigest, UpgradeError> {
 }
 
 impl Base64EncodedDigest {
-    #[inline(always)]
-    pub fn as_str(&self) -> &str {
-        // Safety: Base64 is guaranteed to be ASCII and therefore, valid UTF-8.
```
papa-zakari, line 33: We're all human. If a regression occurs in the base64 crate, we want to be resilient.

PR #557, opened 2026-05-11
feat(via): Payloadz uses the zeroize crate

papa-zakari:
This PR delegates zeroization logic to the `zeroize` crate. This simplification allows future improvements to the `zeroize` crate to apply to users of the `Payloadz` trait. In other words, it's one less thing for us to worry about.

Review from ChatGPT: https://chatgpt.com/share/6a01f8da-d564-83ea-9446-afa9bfbfe89d

### DIFF
--- a/via/src/request/payload.rs
+++ b/via/src/request/payload.rs
@@ -6,11 +6,10 @@ use serde::de::DeserializeOwned;
 use std::future::Future;
 use std::marker::PhantomData;
 use std::pin::Pin;
-use std::ptr;
 use std::rc::Rc;
-use std::sync::atomic::{Ordering, compiler_fence};
 use std::task::{Context, Poll, ready};
 use std::time::Duration;
+use zeroize::Zeroize;
 
 use crate::{Error, err};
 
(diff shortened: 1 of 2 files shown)


via/src/request/payload.rs
```diff
@@ -208,35 +207,20 @@ fn deserialize_utf8(data: Vec<u8>) -> Result<String, Error> {
...
-///   2. `compiler_fence` is called after each frame is zeroized
-///
-/// [zeroize]: https://crates.io/crates/zeroize/
-unsafe fn unfenced_zeroize(frame: &mut Bytes) {
+/// To safely call this fn, you must guarantee unique access to the buffer that
+/// `Bytes` points to. This can be achieved by calling `Bytes::is_unique`.
+unsafe fn zeroize_bytes(frame: &mut Bytes) {
     let len = frame.remaining();
     let ptr = frame.as_ptr() as *mut u8;
+    let buf = std::ptr::slice_from_raw_parts_mut(ptr, len);
```
papa-zakari, line 217: I didn't think of converting `frame` to a `&mut [u8]` at the time that the zeroization logic was implemented. This should be as efficient and even safer than the previous iteration.

via/src/request/payload.rs
```diff
@@ -350,13 +333,12 @@ impl Payloadz for Aggregate {
...
-                unfenced_zeroize(frame);
+                zeroize_bytes(frame);
             }
-        }
 
-        // Ensures sequential access to the buffers contained in self.
-        // A necessary step after zeroization.
-        release_compiler_fence();
+            // Make the visible length of the frame buffer 0.
+            frame.advance(frame.remaining());
```
papa-zakari, line 340: ChatGPT mentioned that it is likely preferred to calculate the length of the frame before we zeroize it.

I disagree with this comment as we only care about truncating the visible length at the time of truncation.

The length of the frame should not change after zeroization. If it does, we should consider it an anomaly and not panic. 

Truncation is less important after zeroization anyways.

via/src/request/payload.rs
```diff
@@ -375,22 +358,20 @@ impl Payloadz for Aggregate {
...
             // The precondition at the top of this function ensures that we
             // have unique access to self and therefore, can mutate the buffer.
             unsafe {
-                unfenced_zeroize(frame);
+                zeroize_bytes(frame);
             }
 
-            // Ensures sequential access to the buffers contained in self.
-            // A necessary step after zeroization.
-            release_compiler_fence();
```
papa-zakari, line 387: At the time of writing the zeroization methods I didn't think about the specific use cases.

Now, it is apparent that the majority of use cases tend to favor small payloads that consist of a single frame. For example, login, account creation, updating some secret parameter all fit in a single frame.

Even if ordering doesn't matter because of the "fail early" implementation of the uniqueness check, there is likely no difference for valid use cases.

PR #564, opened 2026-06-07
fix(via): continue on EMFILE only

papa-zakari:
Reduces the accept errors in which the server continues to listen for new connections to `EMFILE` on POSIX. This enhances security for applications running in hostile environments.

Anyone running an application in a hostile environment should prefer immutably replacing the node rather than simply restarting the server. If a server becomes compromised, replacing the node immutably makes the most sense. The likelihood of an adversary gaining shell access a second time to an entirely new node should be much lower.

### DIFF
--- a/via/src/server/accept.rs
+++ b/via/src/server/accept.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::TcpListener;
 use tokio::sync::Semaphore;
-use tokio::task::{JoinSet, coop};
+use tokio::task::{self, JoinSet, coop};
 use tokio::time::timeout;
 
 #[cfg(any(feature = "native-tls", feature = "rustls-23"))]
@@ -79,14 +79,10 @@ where
                     log!("error(accept): {}", error);
 
                     #[cfg(unix)]
-                    let Some(12 | 23 | 24) = error.raw_os_error() else {
-                        continue;
-                    };
-
-                    #[cfg(windows)]
-                    let Some(10024 | 10055) = error.raw_os_error() else {
-                        continue;
-                    };
+                    if let Some(24) = error.raw_os_error() { // EMFILE
+                        task::yield_now().await; // We're at capacity.
+                        continue; // Try again after yielding to the runtime.
+                    }
 
                     break ExitCode::FAILURE;
                 }
--- a/via/src/server/accept.rs
+++ b/via/src/server/accept.rs
@@ -77,14 +77,41 @@ where
                 Ok(stream) => stream,
                 Err(error) => {
                     log!("error(accept): {}", error);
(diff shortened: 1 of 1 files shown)


via/src/server/accept.rs
```diff
@@ -79,14 +79,10 @@ where
                     log!("error(accept): {}", error);
 
                     #[cfg(unix)]
-                    let Some(12 | 23 | 24) = error.raw_os_error() else {
-                        continue;
-                    };
-
-                    #[cfg(windows)]
-                    let Some(10024 | 10055) = error.raw_os_error() else {
```
papa-zakari, line 87: Windows is much less likely to run into fd limits due to the nature of it's networking APIs. Sockets are not files.

via/src/server/accept.rs
```diff
@@ -77,18 +77,41 @@ where
...
+                            Some(code @ (12 | 23)) => ExitCode::from(code as u8),
+
+                            // EMFILE
+                            //
+                            // The server is at capacity. Yield to the runtime.
+                            //
+                            // A connection task is likely using an fd that the
+                            // application developer did not account for.
+                            Some(24) => {
+                                tokio::task::yield_now().await;
```
papa-zakari, line 95: It's likely better to yield to the runtime as it's really our only choice if we're unable to accept a connection due to fd limits. We can be sure that the CPU would spin without yielding here.

Exiting would prevent someone from implementing a back off based retry flow for files. This isn't something that we would do for an in-house file server implementation as back off based retry is reentrancy attack surface. However, we don't want to preclude anyone from implementing such a thing.

Any Via application built following the best practices outlined by our guides and documentation is unlikely to take this branch. We believe in deterministic resource usage. That means treating file system I/O as we would any other external service. It becomes a secondary form of limiting concurrency.

For example, resource pools are commonly used to guarantee a fixed amount of connections to a database. This means that at capacity `config.max_connections` must compete for `n` number of handles to a resource in the pool. In this case, a database connection. However, the same logic can be applied to a POSIX file server. We do not use `send_file` to stream files to the client. Therefore, the maximum concurrency of the file system is likely the number of worker processes spawned by the runtime.

The default configuration allows for `13` additional file descriptors to be used to concurrently serve at most `config.max_connections`. This means that if you have a server with more than `13` CPU cores or if you configure the tokio runtime to use more than `13` worker process, you must either adjust `config.max_connections` or `ulimit` to support the additional workers.

We suggest keeping `ulimit -n = 1024`. This is open source software and is provided "as is". Therefore, our users are welcome to do what they feel comfortable with.

If you notice that your application is using less than 100% of the CPU, consider using a smaller node to run the application or scale horizontally on the same node by running multiple processes a la Redis Sentinal or with `SO_REUSEPORT` if you prefer that the kernel decide as opposed to your load balancer.

Determinism is the backbone of high assurance software and we prefer it to a lucky guess every day of the week.

PR #567, opened 2026-06-07
fix(via): unused warning in drain_connections in release

papa-zakari:
Fixes an unused variable warning in release builds now that `log!` is fully removed in release.

### DIFF
--- a/via/src/server/accept.rs
+++ b/via/src/server/accept.rs
@@ -198,13 +198,17 @@ async fn drain_connections(immediate: bool, mut connections: JoinSet<Result<(), 
     );
 
     while let Some(result) = connections.join_next().await {
+        #[cfg(not(debug_assertions))]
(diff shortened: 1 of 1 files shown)


via/src/server/accept.rs
```diff
@@ -198,13 +198,17 @@ async fn drain_connections(immediate: bool, mut connections: JoinSet<Result<(),
...
+        #[cfg(debug_assertions)]
         match result {
             Ok(Ok(_)) => {}
             Err(error) => {
-                log!(error(connection), "{}", error);
+                log!(error(connection), "{}", &error);
             }
             Ok(Err(error)) => {
-                log!(error(service), "{}", error);
+                log!(error(service), "{}", &error);
```
papa-zakari, line 211: Also logs errors by ref rather than by value.

PR #574, opened 2026-06-10
refactor(examples/files): cleanup file server example

papa-zakari:
Cleans up the file server example. In theory, these should be identical in functionality. This version is more refined, concise, and explicit about purpose.

The main difference in functionality is that file server state and config is stored in the global application state rather than fields of a middleware struct. Both should be equally safe. Their both obfuscated structs that are heap allocated with a single layer of dynamic dispatch. Not all that different than a boxed future.

This demonstrates our DI API nicely so it's probably a better example.

The math used to determine resource usage ahead of runtime also makes a lot more sense.

This is also helpful example for anyone looking to use additional resources such as a database pool or http client connection pool with Via.

### DIFF
--- a/examples/files/main.rs
+++ b/examples/files/main.rs
@@ -1,12 +1,13 @@
+#![cfg(feature = "fs")]
+
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 use std::sync::Arc;
 use std::{env, thread};
 use tokio::sync::Semaphore;
-use via::{Error, Server};
-
-#[cfg(feature = "fs")]
-use via::{Next, Request, ResultExt};
+use via::response::File;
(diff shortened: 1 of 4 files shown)


examples/files/main.rs
```diff
@@ -1,21 +1,26 @@
...
+/// calculate the number of file descriptors required for the server to never
+/// trigger an EMFILE.
+const RT_FD_REQUIREMENT: usize = 13;
 
 #[allow(dead_code)]
-struct ServeFrom {
+struct Unicorn {
     /// The directory from which files can be served.
-    public_dir: PathBuf,
+    public_dir: Arc<Path>,
```
papa-zakari: Arc path is a nice improvement for both performance and security. This is arguably the most meaningful change in this PR.

PR #575, opened 2026-06-13
fix(via): streaming a file never finishes

papa-zakari:
Fixes a bug where streaming a file never finishes. This is a regression in performance but it makes the file server example work on devices other than my own. Cargo was likely caching a build that worked and incrementally built the parts that changed.

### DIFF
--- a/examples/files/main.rs
+++ b/examples/files/main.rs
@@ -78,7 +78,7 @@ impl<T: Send + Sync> Middleware<T> for ServeFrom {
                 mime_guess::mime::TEXT_HTML_UTF_8
             };
 
-            let response = File::open(&path)
+            let response = File::open(path)
                 .content_type(&mime_type)
                 .with_last_modified()
                 .permit(permit)
--- a/via/src/response/file.rs
+++ b/via/src/response/file.rs
@@ -56,13 +56,13 @@ struct FileStream {
 /// If the file size is less than `max_alloc_size`, the contents will be
 /// eagerly read into memory.
 ///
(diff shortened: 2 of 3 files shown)


via/src/response/file.rs
```diff
@@ -63,7 +63,7 @@ async fn open(path: &Path, max_alloc_size: u64) -> Result<Open, Error> {
 
     if metadata.len() > max_alloc_size {
         let mut file = TokioFile::open(path).await.map_err(Error::from_io_error)?;
-        file.set_max_buf_size(0); // Buffering occurs in ReaderStream.
+        file.set_max_buf_size(MAX_BUFFER_SIZE);
```
papa-zakari: As part of our goal of never allowing hidden buffering in Via, we'll explore an option that involves 2 buffers instead of 3. The `tokio::fs::File` implementation has an internal buffer that we do not control.

examples/files/main.rs
```diff
@@ -78,7 +78,7 @@ impl<T: Send + Sync> Middleware<T> for ServeFrom {
                 mime_guess::mime::TEXT_HTML_UTF_8
             };
 
-            let response = File::open(&path)
+            let response = File::open(path)
```
papa-zakari, line 81: We take ownership of the path as it enforces a clear ownership boundary in the API and prevents holding a reference to path for longer than necessary.

via/src/response/file.rs
papa-zakari: Memcpy is an excellent source of noise. I think it's probably worth waiting for the community to have an opinion on this.

We can try to guess as much as we want but real world experience always trumps speculation.

PR #578, opened 2026-06-13
fix(via): unused reexport when tls backends are enabled

papa-zakari:
Fixes compiler warnings when a TLS backend is enabled.

### DIFF
--- a/examples/rustls/main.rs
+++ b/examples/rustls/main.rs
@@ -12,6 +12,27 @@ async fn hello(request: Request, _: Next) -> via::Result {
     Response::build().text(format!("Hello, {}! (via TLS)", name))
 }
 
+#[cfg(all(
+    any(feature = "aws-lc-rs", feature = "ring"),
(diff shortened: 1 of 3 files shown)


examples/rustls/main.rs
```diff
@@ -12,6 +12,27 @@ async fn hello(request: Request, _: Next) -> via::Result {
     Response::build().text(format!("Hello, {}! (via TLS)", name))
 }
 
+#[cfg(all(
+    any(feature = "aws-lc-rs", feature = "ring"),
+    any(feature = "tokio-tungstenite", feature = "tokio-websockets")
+))]
+async fn echo(mut channel: via::ws::Channel, _: via::ws::Request) -> via::ws::Result {
```
papa-zakari, line 19: This makes it easy to test websockets with a TLS backend. Anyone who is considering using Via for a secure messaging application will benefit from this.

PR #581, opened 2026-06-15
docs(via): document public members in mod request

papa-zakari:
Documents the public members of the top-level request module.

### DIFF
--- a/via/src/request/mod.rs
+++ b/via/src/request/mod.rs
@@ -376,8 +376,7 @@ impl<App> Request<App> {
         (Coalesce::new(self.body), self.app)
     }
 
-    /// Consumes the request and returns a tuple containing it's parts.
-    ///
+    /// Consumes the request and returns a tuple containing its parts.
     #[inline]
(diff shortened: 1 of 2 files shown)


via/src/request/mod.rs
```diff
@@ -18,12 +18,157 @@ use crate::error::Error;
...
+///     };
+///
+///     // Spawn a detached task that takes ownership of the request
+///     // envelope, a cloned copy of the active user, and a shared
+///     // handle to your application.
+///     tokio::spawn({
+///         // Prepare the telemetry task dependencies.
+///         let action = Action::Login {
+///             request: envelope,
+///             user: me.clone(),
```
papa-zakari, line 89: This clone is an unnecessary allocation but I like the way that the response is built after spawning the task. That's part of the purpose of this API.

Another use-case is discarding the body and retaining the request for a long-lived connection (web sockets). 

This use-case is more interesting and justifies making the API public beyond our need for retaining the envelope in the request struct passed to web socket handlers.

Another good example could demonstrate how to use  `http-body-util` with a `via::Request`. However, that is boring by comparison.

PR #584, opened 2026-06-17
docs(via): document the middleware trait 

papa-zakari:
Documents the `Middleware` trait. Providing guidance and patterns on how users can compose middleware to build an application.

### DIFF
--- a/via/src/middleware.rs
+++ b/via/src/middleware.rs
@@ -41,13 +41,17 @@ pub type Result<T = Response> = std::result::Result<T, Error>;
 ///
 /// async fn hello(request: Request, _: Next) -> via::Result {
 ///     //                           ^^^^^^^
-///     // The `next` argument is ignored. This is a terminal middleware.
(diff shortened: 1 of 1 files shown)


via/src/middleware.rs
```diff
@@ -17,7 +17,232 @@ pub type BoxFuture<T = Result> = Pin<Box<dyn Future<Output = T> + Send>>;
...
+/// discrete, explicit impl for a concrete type.
+///
+/// ## Higher-Order Middleware
+///
+/// Higher-order middleware modifies how subsequent middleware behaves or
+/// decorates the response built by a terminal middleware.
+///
+/// Examples of higher-order middleware include:
+///
+/// - Authentication middleware that decides whether a request can continue.
```
papa-zakari: Usually I don't include punctuation in bulleted lists. However, these are full-blown run-on sentences.

papa-zakari: I changed my mind

PR #588, opened 2026-06-19
docs(via): document the predicate trait

papa-zakari:
Documents the `Predicate` trait.


via/src/guard/method.rs
```diff
@@ -1,11 +1,21 @@
 //! Match on a method.
```
papa-zakari: I'm probably going to play around with the names here and then split this into a separate pull request. Method guards are useful for filters despite some duplication with the router switch.

I don't want to use guards for routing as it's an entirely separate concern (or layer) of the framework. Guards are stateless, higher-order middleware where a switch is a decision tree for a terminal middleware.

PR #607, opened 2026-07-06
feat(via): capability-bounded rescue policies

papa-zakari:
Refactors the error rescue middleware to use capability-bounded recovery policies implemented with a sealed trait. The rescue middleware is often times the first line of defence against adversarial connections. This refactor turns what could otherwise be a high value exploit into noisy framework code.

### DIFF
--- a/via/src/error/rescue.rs
+++ b/via/src/error/rescue.rs
@@ -1,156 +1,294 @@
+//! Recover from service errors with capability-bounded policies.
+//!
+//! Unlike traditional exception handlers or recovery callbacks, rescue
+//! policies cannot arbitrarily construct responses. Instead, they observe a
+//! recovered [`Error`] and request predefined rendering behavior such as JSON
+//! output, canonical reason phrases, or connection termination.
+//!
+//! This separation of concerns keeps response construction within Via while
+//! still allowing applications to customize how errors are presented and
+//! reported. Policies may also perform side effects such as logging or
+//! telemetry without influencing the rendered response.
+//!
+//! Policies compose declaratively:
+//!
+//! ```no_run
+//! use via::error::rescue;
+//!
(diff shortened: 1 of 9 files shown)


via/src/error/rescue.rs
```diff
@@ -1,156 +1,294 @@
...
+}
+
+struct Render<'a> {
+    flags: &'a PolicyMask,
+    error: &'a Error,
+}
+
+bitflags! {
+    #[derive(Clone, Copy, Eq, PartialEq)]
+    struct PolicyMask: u8 {
```
papa-zakari, line 116: These flags are noise for `OPTIONS`, `PATCH`, and `DELETE` requests.

`GET` and `POST` can instead be easily confused with the bit masks in the `tokio::select!` arms of the accept loop and connection task.

The frequency of errors should be lower that turns of the accept loop. Making their cadence align with the aforementioned request methods as well.  

*tldr;*

Bit masks are not a good signal for adversaries that wish to remain undetected.

via/src/error/rescue.rs
```diff
@@ -1,156 +1,294 @@
...
+///
+/// This trait is sealed and cannot be implemented outside of Via.
+pub trait Policy: sealed::Sealed {
+    fn apply(&self, error: &Error, flags: PolicyFlags) -> PolicyFlags;
+}
+
+/// Compose two recovery policies.
+///
+/// Policies are applied from left to right.
+pub struct And<T, U>(T, U);
```
papa-zakari, line 58: For this use case, a bounded yet nested policy structure with `And` creates structural indirection that makes blind patching impossible.

If someone wishes to compromise the inspect hook of your error middleware, they have to study your app and pray that the closure you pass to inspect does not get inlined.

PR #610, opened 2026-07-07
feat(via): via::before is #[inline(always)]

papa-zakari:
Unconditionally inlines the constructor of  `via::Before` . We don't want to do binary patchers any favors.

### DIFF
--- a/via/src/before.rs
+++ b/via/src/before.rs
@@ -182,6 +182,7 @@ pub struct Before<T, U> {
 /// }
 /// ```
 /// </details>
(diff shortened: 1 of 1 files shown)


via/src/before.rs
```diff
@@ -182,6 +182,7 @@ pub struct Before<T, U> {
 /// }
 /// ```
 /// </details>
+#[inline(always)]
```
papa-zakari, line 185: This doesn't have to be macro because of the implementation and arity.

PR #612, opened 2026-07-12
feat(via): router scope ensures dsl correctness

papa-zakari:
Enforces router DSL correctness by requiring routes to define middleware before defining descendants or assigning a terminal middleware ("responder").

This change requires that the router be constructed separately from the application in order to properly support middleware and terminal middleware at the application root "/".

The result is actually pretty nice considering that it cleanly separates DI from route definitions. It also allows routes to be defined before setting up application dependencies.It's a best practice to connect to the database as late as possible but just before the server starts accepting connections.

The chat app will be update immediately so everyone has an example to follow. These are last minute growing pains right before a major release.

### DIFF
--- a/examples/cookies.rs
+++ b/examples/cookies.rs
@@ -1,13 +1,13 @@
 use cookie::{Cookie, Key, SameSite};
 use std::process::ExitCode;
 use std::time::Duration;
-use via::{Error, Next, Request, Response, ResultExt, Server};
+use via::{Error, Next, Request, Response, ResultExt, Router, Server, cookies};
 
-struct Unicorn {
+struct Buttercup {
     secret: Key,
 }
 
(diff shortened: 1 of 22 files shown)


examples/cookies.rs
```diff
@@ -1,13 +1,13 @@
 use cookie::{Cookie, Key, SameSite};
 use std::process::ExitCode;
 use std::time::Duration;
-use via::{Error, Next, Request, Response, ResultExt, Server};
+use via::{Error, Next, Request, Response, ResultExt, Router, Server, cookies};
 
-struct Unicorn {
+struct Buttercup {
```
papa-zakari, line 6: The cutesy names help keep Via approachable while also being a bit tongue in cheek regarding our high assurance status.

This a little bit "tough guys wear pink" and mostly "we want female soldiers to write code with this with a callback to childhood to maintain their humanity while at war".

🦄

PR #620, opened 2026-07-20
fix(via): let _ = arg; when log! has fmt args

papa-zakari:
Updates the log macro to ignore arguments in release builds.


papa-zakari:
This can inadvertently cause

PR #630, opened 2026-07-23
docs(via): safety comment on impl Send for Facade

papa-zakari:
Includes a long overdue safety comment on the `unsafe impl Send for Facade {}`. As always—if you see a soundness issue, call it out.


papa-zakari:
Closed in favor of #631.

PR #632, opened 2026-07-24
feat(via-pubsub): bounded pipeline size

papa-zakari:
Creates an upper bound on the number of commands that can be included in a redis pipeline by dividing a global maximum 64kb by a configurable max_event_size. This allow us to determine the pathological worst case without keeping a running total. We are high assurance. We assume worst case.

### DIFF
--- a/examples/chat/src/main.rs
+++ b/examples/chat/src/main.rs
@@ -30,6 +30,9 @@ use routes::auth::{login, logout, me};
 use routes::{channels, reactions, threads, users};
 use util::session::{self, auth_required, authenticate};
 
+#[cfg(any(feature = "tokio-tungstenite", feature = "tokio-websockets"))]
+use crate::app::MAX_EVENT_SIZE;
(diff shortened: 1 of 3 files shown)


examples/chat/src/main.rs
```diff
@@ -89,7 +92,12 @@ fn routes(home: via::Route<Unicorn>) {
 
     // The /api/chat route.
     #[cfg(any(feature = "tokio-tungstenite", feature = "tokio-websockets"))]
-    path.route("/chat", via::get(authenticate(via::ws(routes::chat))));
+    path.route(
+        "/chat",
+        via::get(authenticate(
+            via::ws(routes::chat).max_message_size(Some(MAX_EVENT_SIZE)),
```
papa-zakari, line 98: If `max_message_size` is not `MAX_EVENT_SIZE`, buffering occurs in the outbound pubsub channel.

PR #634, opened 2026-07-27
feat(examples/chat): new users must confirm password

papa-zakari:
Updates the chat example to require a `confirmPassword` field for new users who are looking to create an account.

### DIFF
--- a/examples/chat/src/models/user.rs
+++ b/examples/chat/src/models/user.rs
@@ -1,4 +1,4 @@
-use argon2::PasswordHash;
+use argon2::{Argon2, PasswordHash, PasswordVerifier};
 use diesel::deserialize::{self, FromSql, FromSqlRow};
 use diesel::helper_types::{AsSelect, InnerJoin, InnerJoinOn, Select};
 use diesel::pg::{Pg, PgValue};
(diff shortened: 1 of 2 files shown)


examples/chat/src/models/user.rs
```diff
@@ -127,34 +129,58 @@ where
...
+    pub async fn create(connection: &mut Connection<'_>, mut init: NewUser) -> via::Result<Self> {
+        // Parse a `PasswordHash` from the PHC string created when the
+        // `NewUser` was deserialized.
+        let password_hash = init.password.hash()?;
+
+        // Verify the "confirmPassword" field against the password hash.
+        let verify_result = verify_password(init.confirm_password.as_bytes(), &password_hash);
+
+        // Eagerly zeroize the  plaintext "confirmPassword" field.
+        init.confirm_password.zeroize();
```
papa-zakari, line 170: The lifetime of the plaintext `confirm_password` is that of the parsed password hash.

Neither the parsed PHC string or the plaintext `confirm_password` field are accessible while waiting for the insert to complete.

This helps ensure that passwords and secrets have minimum residency especially when they exist in memory as plaintext.

examples/chat/src/models/user.rs
```diff
@@ -127,34 +129,60 @@ where
...
             .filter(by_email(&params.email))
             .first_async::<(User, Password)>(connection)
             .await
             .or(Err(Unauthorized))?;
 
-        Argon2::default()
-            .verify_password(params.password.as_bytes(), &password.hash()?)
-            .map_or_else(|_| Err(Unauthorized.into()), |_| Ok(user))
+        // Parse a `PasswordHash` from `password`.
+        let password_hash = password.hash()?;
```
papa-zakari, line 153: No reason to put the password hash in a separate block as it does not shrink the it's lifetime.

As soon as verification is performed, a result is returned.

The `await` happens before the password is confirmed during authentication.

PR #635, opened 2026-07-28
feat(via-diesel): keyset pagination

papa-zakari:
Adds a keyset pagination API to via-diesel. The API works similar to other pagination APIs with an additional step to bind the keyset columns to the values deserialized from the URI query.

Keyset pagination works with any database backend and any query that implements the `LimitDsl` and `BoxedDsl` with the columns to support the filter expressions required by the keyset.

Keysets require two columns. A pivot column and a tiebreaker column. Neither can be null. Typically these are an automatically generated timestamp column (`created_at`) and a primary key.

The keyset pagination API can be seen in action in the threads and replies resource of the chat example.

```rust
async fn index(request: Request, _: Next) -> via::Result {
    // Get the channel id from the subscription we loaded during authorization.
    // If the current user does not have an active subscription, 404 Not Found.
    let channel_id = request.channel_id().or_not_found()?;

    // Source keyset arguments from the URI query.
    // If the limit query param is > the default per page, 400 Bad Request.
    let by_keyset = request.query::<Keyset<Iso8601, Id, PER_PAGE>>()?;

    // Load a page of threads.
    let mut feed = {
        // Checkout a database connection.
        let mut connection = request.app().database().await?;

        // Load the threads in the channel, paginated `by_keyset`.
        let threads = ThreadWithUser::query()
            .filter(by_channel(channel_id).and(is_thread()))
            .order(recent())
            .page(by_keyset.of(threads::created_at, threads::id))
            .load_async(&mut connection)
            .await?;

        // Side load the reactions to the threads in `threads`.
        Reaction::to_threads(&mut connection, threads).await?
    };

    feed.reverse(); // Presented as append-only.

    Response::build().data(feed)
}
```


via-diesel/src/paginate/keyset.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::fmt::{self, Display, Formatter};
+use std::str::FromStr;
+use time::{OffsetDateTime, format_description::well_known::Rfc3339};
+use via::request::QueryParams;
+
+use super::{Limit, Paginate};
+use crate::id::{self, Id};
+
+type DtzValueExpr = <OffsetDateTime as AsExpression<sql_types::Timestamptz>>::Expression;
+type PkValueExpr = <Id as AsExpression<id::SqlType>>::Expression;
```
papa-zakari: This is the main blocker. I want to generalize `Id` so that way, users don't have to use `via_diesel::Id` and the bounds on the impl of `Pagination` will require that the left and right hand side of the tie breaker expr have the same `sql_type`.

via-diesel/src/paginate/keyset.rs
```diff
@@ -0,0 +1,186 @@
...
+use diesel::{Expression, QueryDsl, sql_types};
+use std::fmt::{self, Display, Formatter};
+use std::str::FromStr;
+use time::{OffsetDateTime, format_description::well_known::Rfc3339};
+use via::request::QueryParams;
+
+use super::{Limit, Paginate};
+use crate::id::{self, Id};
+
+type DtzValueExpr = <OffsetDateTime as AsExpression<sql_types::Timestamptz>>::Expression;
```
papa-zakari: Chrono should be supported as well.

examples/chat/src/main.rs
```diff
@@ -128,8 +128,8 @@ fn routes(home: via::Route<Unicorn>) {
                         .route("/:reaction-id", reactions::member());
 
                     // The ./:thread-id/replies resource
-                    path.route("/replies", threads::collection())
-                        .route("/:reply-id", threads::member())
+                    path.route("/replies", replies::collection())
```
papa-zakari, line 131: Using the same resource for threads and replies is a pretty neat trick.

However, it's best to avoid the added polymorphism in our flagship example of a high assurance chat API.

PR #638, opened 2026-08-04
feat(via): lag is an event not an error

papa-zakari:
Introduces a lag event as part of via-pubsub and the necessary APIs in via required to support first class lag semantics.

### DIFF
--- a/examples/chat/src/routes/chat.rs
+++ b/examples/chat/src/routes/chat.rs
@@ -190,11 +190,11 @@ fn deserialize_client_event(message: &Message) -> via::Result<ClientEvent> {
 fn serialize_lag_notification(length: u64) -> ws::Result<String> {
     #[derive(Serialize)]
     #[serde(content = "data", rename_all = "lowercase", tag = "type")]
-    enum LagNotifiction {
+    enum LagNotification {
         Lag { length: u64 },
     }
 
-    serde_json::to_string(&LagNotifiction::Lag { length }).or_continue()
+    serde_json::to_string(&LagNotification::Lag { length }).or_continue()
 }
 
 #[cfg(all(debug_assertions, feature = "tokio-tungstenite"))]
--- a/examples/chat/src/routes/chat.rs
+++ b/examples/chat/src/routes/chat.rs
@@ -1,4 +1,4 @@
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use tokio::task::coop;
 use via::deny;
 use via::error::{Catch, Propagate};
@@ -139,6 +139,13 @@ pub async fn chat(mut channel: Channel, request: Request) -> ws::Result {
 
         if let Some(event) = inbound {
             match event {
+                // Lag detected in `subscription`.
+                PeerEvent::Lag(len) => {
+                    log!(info(chat = 1), "lag notification; len = {}", len);
+                    channel.send(serialize_lag_notification(len)?).await?;
+                    return ws::restart().await;
+                }
+
                 // The user logged out.
                 PeerEvent::Logout => {
                     log!(info(chat = 1), "ws session ended");
@@ -147,8 +154,7 @@ pub async fn chat(mut channel: Channel, request: Request) -> ws::Result {
 
                 // Notification received from a peer.
                 PeerEvent::Relay(notification) => {
-                    log!(info(chat = 1), "relay notification");
-                    channel.send(notification).await?
+                    channel.send(notification).await?;
                 }
 
                 // The user was invited to a channel.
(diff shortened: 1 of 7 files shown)


examples/chat/src/routes/chat.rs
```diff
@@ -139,6 +139,13 @@ pub async fn chat(mut channel: Channel, request: Request) -> ws::Result {
 
         if let Some(event) = inbound {
             match event {
+                // Lag detected in `subscription`.
+                PeerEvent::Lag(len) => {
+                    log!(info(chat = 1), "lag notification; len = {}", len);
+                    channel.send(serialize_lag_notification(len)?).await?;
+                    return ws::restart().await;
```
papa-zakari, line 146: Notify the client of the lag and restart the listener.

This rebuilds the users subscription while keeping the underlying connection in tact. While waiting, the client can chose to synchronize with the canonical state.

examples/chat/src/routes/chat.rs
```diff
@@ -181,6 +187,16 @@ fn deserialize_client_event(message: &Message) -> via::Result<ClientEvent> {
     Ok(serde_json::from_str(text)?)
 }
 
+fn serialize_lag_notification(length: u64) -> ws::Result<String> {
+    #[derive(Serialize)]
+    #[serde(content = "data", rename_all = "lowercase", tag = "type")]
+    enum LagNotifiction {
```
papa-zakari: The shape of lag notifications are determined by the implementation.

Length can be beneficial for non-linear data presentations where the lag length / number of entities on screen provides a level-of-confidence while a chat application may chose to treat lag as a boolean.

via/src/error/mod.rs
```diff
@@ -44,13 +44,15 @@ enum ErrorSource {
     Message(String),
     Other(BoxError),
     Json(serde_json::Error),
+    Restart,
```
papa-zakari, line 50: In the eyes of the websocket reactor (`Run`), a lag error is a special type of error that does not allocate or hold any contextual data.

via/src/error/mod.rs
```diff
@@ -251,6 +267,7 @@ impl From<Error> for BoxError {
             ErrorSource::Message(string) => string.into(),
             ErrorSource::Other(source) => source,
             ErrorSource::Json(source) => source.into(),
+            ErrorSource::Restart => "restart".to_owned().into(),
```
papa-zakari: This is effectively unreachable.

papa-zakari: Going to mark this with `cold_path`.

via/src/ws/mod.rs
```diff
@@ -68,3 +70,15 @@ where
...
     Ws::new(listener)
 }
+
+/// Explicitly restarts a listener while keeping the connection intact.
+///
+/// The existing listener will first yield so the reactor can complete pending
+/// tasks.
+pub async fn restart() -> Result {
+    // Immediately yield so the reactor can complete any pending tasks.
+    tokio::task::yield_now().await;
```
papa-zakari, line 80: The listener must yield unconditionally in order for the reactor to finish any pending operations.

Otherwise, additional work may be scheduled if the task has the budget for it.

via/src/ws/run.rs
```diff
@@ -315,12 +315,18 @@ where
...
+        let mut poll = match this.facade.as_mut() {
+            Some(facade) => Pin::new(facade).poll(context),
+            None => Pin::new(this.reconnect()).poll(context),
         };
 
-        match Pin::new(future).poll(context) {
+        if let Poll::Ready(Err(ControlFlow::Continue(ref error))) = poll {
+            if error.is_restart() {
+                poll = Pin::new(this.reconnect()).poll(context);
+            }
```
papa-zakari: If an error occurred due to an explicit restart request, replace the future and poll it immediately.

There is likely budget for it since well-structured listeners try to drain the subscription before reading from socket and lag errors propagate immediately.

via/src/ws/run.rs
```diff
@@ -333,9 +339,13 @@ where
 
                 log!(warn(ws = 1), "{}", &error);
 
-                this.reconnect();
```
papa-zakari, line 336: `Run::reconnect` is never used as a side effect.

examples/chat/src/routes/chat.rs
papa-zakari: Also, it's worth mentioning that the lag length includes notifications that are filtered by the subscription. Therefore, it can be used as an approximation but it assumes that no messages were filtered. 

Totally fine because it errors on the side of caution. In practice this value is likely `1..concurrency` and will only be used by applications that also show more than 3 decimal places when displaying floats.

I have a hunch that this will be useful for realtime applications that plot an updating location on a map.

